PR #2, opened 2021-05-03
Add Spectral version 2

adriancarriger:


src/testing.ts
```diff
@@ -90,36 +90,44 @@ export const credentials = {
 export const loggerMock = (): ActionLogger => ({
   debug: (spyOn(console, "debug") as unknown) as ActionLoggerFunction,
   info: (spyOn(console, "info") as unknown) as ActionLoggerFunction,
-  log: (spyOn(console, "log") as unknown) as ActionLoggerFunction,
```
taylorreece, line 93: We'll want to add this one back in.

taylorreece (commented):
Do we want to maybe keep a longer-living "v2" branch distinct from "main" for a while while we test these changes? When we do merge to main, we'll want to coordinate updating of docs and whatnot, so users who pull latest aren't surprised by syntax changes.

src/testing.ts
adriancarriger: Fixed: https://github.com/prismatic-io/spectral/compare/8cf1d2e74540c796193db6fe8f9a00815f6c8b95..bcede00802cf0b2bcc3f519d1b94baee32379acc

taylorreece (approved):

PR #3, opened 2021-05-18
Expand type coercion and allow for default values to be specified for util functions

taylorreece:
Would this do the trick to allow component authors to handle `unknown` inputs better?

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -23,10 +23,7 @@ const toBool = (value: unknown, defaultValue?: boolean) => {
   }
 
   if (typeof value === "undefined") {
-    if (typeof defaultValue === "undefined") {
-      return true;
-    }
-    return defaultValue;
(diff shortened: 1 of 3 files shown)


src/util.ts
```diff
@@ -22,12 +22,19 @@ const toBool = (value: unknown) => {
     }
   }
 
+  if (typeof value === "undefined") {
+    if (typeof defaultValue === "undefined") {
+      return true;
```
adriancarriger, line 27: I'm trying to understand why we return `true` if the value is `undefined`? Wouldn't we just want to call `Boolean` on the value?

adriancarriger (approved):
Thanks for adding these util functions 👍

adriancarriger (approved):
Thanks, latest looks good 👍

PR #4, opened 2021-05-25
update toString to handle number, bool, etc...

kaleb-tringale:


src/util.ts
```diff
@@ -151,21 +152,8 @@ const toData = (value: unknown): DataPayload => {
...
-    return defaultValue || "";
-  }
-
-  if (typeof value === "object" && value !== null) {
-    return value.toString();
-  }
-
-  throw new Error(`Value '${value}' cannot be converted to a String.`);
-};
+const toString = (value: unknown, defaultValue = ""): string =>
```
adriancarriger, line 155: I think you can remove `: string`.

adriancarriger (approved):
Nice!

PR #6, opened 2021-05-27
Removed the conditional expression check. Our terms can be more than …

jakethagle:
…strings so it seemed to be less useful with that in mind


taylorreece:
If this is something we want to get out the door, please bump version in package.json as part of this PR.  Thanks!

taylorreece (approved):
LGTM; thanks!

hipplej (approved):

PR #8, opened 2021-06-01
Kt/update key value list to handle undefined

kaleb-tringale:


src/util.test.ts
```diff
@@ -360,4 +359,30 @@ describe("util", () => {
       );
     });
   });
+  //TODO add an arbitrary for KeyValueList to test unique values
```
hipplej, line 362: Still valid?

hipplej (commented):
Perhaps one stray comment but otherwise LGTM.

PR #12, opened 2021-06-10
Account for empty strings in toBool() and toInt()

taylorreece:
Reasoning: `Boolean('') === false`, so `toBool('')` should probably also return `false` (rather than its current behavior of throwing an error.

For `toInt()`, I'd expect empty string to behave the same as `undefined`.

### DIFF
--- a/src/util.ts
+++ b/src/util.ts
@@ -22,11 +22,7 @@ const toBool = (value: unknown, defaultValue?: boolean) => {
     }
   }
 
-  if (typeof value === "undefined" || value === "") {
-    return Boolean(defaultValue);
-  }
-
-  throw new Error(`Value '${value}' cannot be coerced to bool.`);
+  return Boolean(value || defaultValue);
 };
 
 const isInt = (value: unknown): value is number => Number.isInteger(value);
--- a/src/util.ts
+++ b/src/util.ts
@@ -22,7 +22,7 @@ const toBool = (value: unknown, defaultValue?: boolean) => {
     }
   }
 
-  if (typeof value === "undefined") {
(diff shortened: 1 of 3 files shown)


src/util.ts
```diff
@@ -22,7 +22,7 @@ const toBool = (value: unknown, defaultValue?: boolean) => {
     }
   }
 
-  if (typeof value === "undefined") {
+  if (typeof value === "undefined" || value === "") {
```
adriancarriger, line 25: Would it make more sense to always just `return Boolean(defaultValue);` and not throw an error?

adriancarriger (approved):

src/util.ts
adriancarriger: Yeah, I like that!

src/util.ts
```diff
@@ -41,7 +37,7 @@ const toInt = (value: unknown, defaultValue?: number) => {
...
```
adriancarriger, line 40: I'm wondering if we should just call `Number` at some point instead of throwing (on line 44)

taylorreece: What behavior would you expect if if they pass in a `Number` of 5.5? Would you expect to cast that number to an int (so return `5`?

adriancarriger: It makes me wonder if we should rename the function (or create a new one that is called `toNumber`). Do we really need an int in most cases?

taylorreece: That's a good point. There are definitely times where you want an _int_ but having a `toNumber` is probably a good idea, too. I'll write that up quick.

adriancarriger: Nice, thanks!

PR #13, opened 2021-06-17
Add inline documentation and restructure a few files

taylorreece:
- Add typedoc to this project, so generation of docs is not a project with different typescript version, etc.
- Add inline documentation so we can generate typedocs to be ingested into prismatic.io/docs.
 - Deduplicate some type definitions in server-types.ts that were identical to types defined elsewhere.
 - Rename a few types in server-types.ts to avoid name confusion.
 - Fix a few linting issues where we were using `any`, etc.

I tested these changes against all built-in components, and all built and passed unit tests.


.gitignore
```diff
@@ -105,3 +105,8 @@ dist
 
 # Node packages
 lib/
+
+# Docs
+docs/
+sidebars.js
+sidebars.jse
```
adriancarriger, line 112: Should there be a new line here?

src/types/server-types.ts
```diff
@@ -139,47 +56,23 @@ export interface ActionDefinition {
...
-  stepId: string;
+    | ServerPerformDataStructureReturn
+    | ServerPerformBranchingDataStructureReturn;
 }
 
 /** Collection of input parameters provided by the user or previous steps' outputs */
 interface ActionInputParameters {
-  [key: string]: any;
+  [key: string]:
+    | boolean
```
adriancarriger, line 66: I think if you include `unknown` in a union like this it just compiles down to `unknown`. That means we could just change this to

```typescript
interface ActionInputParameters {
  [key: string]: unknown;
}
```

src/util.ts
```diff
@@ -63,23 +136,49 @@ const toNumber = (value: unknown, defaultValue?: number) => {
...
+ * @returns This function returns the bigint representation of `value`.
+ */
+const toBigInt = (value: unknown): BigInt => {
   if (isBigInt(value)) {
     return value;
   }
 
   try {
-    return BigInt(value);
+    return BigInt(value as string);
```
adriancarriger, line 162: Can we remove the casting here?

adriancarriger (commented):
I really like the improved docs here! 🎉 🎉

src/types/server-types.ts
taylorreece: Good call!

PR #14, opened 2021-06-21
Optional Component Category

hipplej:
Add an optional category to the component definition.

### DIFF
--- a/src/server-types.ts
+++ b/src/server-types.ts
@@ -109,6 +109,8 @@ interface DisplayDefinition {
   label: string;
   /** Description to display to the user. */
   description: string;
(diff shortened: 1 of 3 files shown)


src/server-types.ts
```diff
@@ -109,6 +109,8 @@ interface DisplayDefinition {
   label: string;
   /** Description to display to the user. */
   description: string;
+  /** Category of the Component. */
+  category?: string;
```
taylorreece, line 113: Might a component have multiple categories?

hipplej: I asked the same thing and was told no.

PR #16, opened 2021-07-27
Make category required for public components

adriancarriger:


taylorreece:
Hey @adriancarriger - this is awesome! Could you please make `documentationUrl` a required property for public components (so we don't forget that for built-in Prismatic components), as well?

taylorreece:
We'll also probably want to bump spectral version to 2.1.1 - thanks!

PR #17, opened 2021-08-05
Move authorization config from component to action

ryanwersal:
This removes the component level authorization block in favor of adding it to each action to allow each to specify their own configuration. This includes a major version bump since this completely removes the component authorization block.


taylorreece:
Looks good to me - Thanks @ryanwersal !

PR #18, opened 2021-08-09
Support custom headers in OAuth2 credentials

hipplej:
Needed for upcoming demo, and probably just a good thing to have in general.

[ch7652]


taylorreece (approved):
LGTM, thanks!

PR #19, opened 2021-08-12
Rename isData/toData to improve clarity on their purpose

ryanwersal:
Recent feedback showed that `isData` and `toData` are not very clear over their purpose. I opted to extend the name to be more verbose and hopefully more clear: `isBufferDataPayload` and `toBufferDataPayload`. I'm definitely open to alternatives on the naming as well. I left the old names in and marked them as deprecated.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "3.0.1",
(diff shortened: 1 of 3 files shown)


hipplej (approved):
Seems reasonable to me.

PR #21, opened 2021-08-19
Added isJSON To Util

kaleb-tringale:


taylorreece (commented):
This seems reasonable to me. Please add some documentation with examples to the top similar to the other `is*` functions.

I noticed `isJSON(5)` returns true. I _think_ that's fine, since `5` is technically [valid JSON](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Objects/JSON#other_notes).

taylorreece (approved):
Looks great now - thanks!

PR #23, opened 2021-09-10
Copied over the updated types from conditonal-logic to spectral

jakethagle:
Currently we can only mirror these types by hard coding them. Until we decide to publish conditional logic spectral cant depend on it.


hipplej:
I think we'd normally want to bump the version for this, but we have a few other changes going out anyway so I think this is fine as is.

hipplej (approved):

PR #24, opened 2021-09-13
Add support for custom triggers

hipplej:


src/index.ts
```diff
@@ -66,23 +105,34 @@ const convertAction = (
...
+      TriggerDefinition<any, boolean, TriggerResult<boolean>>
     >;
   }
 ): Component<T> => ({
   version: "placeholder", // Placeholder until we deprecate version in component definitions
   ...definition,
   documentationUrl: definition.documentationUrl || null,
   actions: Object.fromEntries(
-    Object.entries(definition.actions).map(([actionKey, action]) => [
+    Object.entries(definition?.actions || {}).map(([actionKey, action]) => [
```
bkegley, line 125: nit: I think `definition` is always defined so you don't need the optional chaining here.

bkegley (approved):
LGTM!

src/index.ts
hipplej: Thanks.

PR #26, opened 2021-09-29
Export some useful non-templated types

hipplej:
Not sure if this is the best way to export these or not. We want to avoid collisions with existing templated types, which are named the same thing. So this ends up getting used like:
```
import { serverTypes } from "@prismatic-io/spectral";
...
const foo: serverTypes.InputFieldType = "string"; // contrived example
```

Is that what we want?


adriancarriger (approved):
Looks good to me 👍

PR #35, opened 2021-11-08
A flow can have multiple API keys

taylorreece:
webhookApiKeys should allow for `string[]` instead of just `string`.


hipplej (approved):
Bah, I should have remembered that.

PR #37, opened 2021-11-23
Replace Credentials with Connections

ryanwersal:


.gitignore
```diff
@@ -1,3 +1,6 @@
+.rush
```
taylorreece, line 1: Is this necessary if we use `yarn` for Spectral?

ryanwersal: It is not necessary. It's an artifact of how I was working on this changeset and I wasn't sure if it was worth including (potentially ignoring other IDE/tool directories such as `.idea` or similar). Happy to remove that if we want to keep the `.gitignore` cleaner 👍

src/types/Inputs.ts
```diff
@@ -28,11 +30,45 @@ export interface DefaultInputFieldDefinition {
...
+
+export interface Connection<
+  TField extends ConnectionFieldDefinition = ConnectionFieldDefinition
+> {
+  /** Key of the Connection type. */
+  key: TField["connectionKey"];
+  /** Identifier for the Config Variable hosting this Connection. */
+  instanceConfigVarId: string;
+  /** Field values supplied to this Connection. */
+  fields: { [Property in keyof TField["inputs"]]: unknown };
```
taylorreece, line 66: Will a field value ever _not_ be a `string`? Should we type this `: string` instead of `: unkonwn`?

src/types/Inputs.ts
```diff
@@ -28,11 +26,28 @@ export interface DefaultInputFieldDefinition {
...
+export interface CodeInputFieldDefinition extends BaseInputFieldDefinition {
+  type: Extract<InputFieldType, "code">;
   language?: string;
 }
 
+export interface Connection {
+  /** Key of the Connection type. */
+  key: string;
+  /** Identifier for the Config Variable hosting this Connection. */
+  instanceConfigVarId: string;
```
hipplej, line 44: This seems like our implementation details are leaking.

hipplej (approved):
I'm concerned about the `instanceConfigVarId` thing but it seems like there's probably a reason for it that just isn't clear to me.

PR #41, opened 2022-01-19
Document and clean up SOAP client types and utility functions

taylorreece:
- Document SOAP client types and utility functions
- Clean up some function overloading
- Ensure SOAP components can be tested on Windows systems

### DIFF
--- a/src/clients/soap/utils.ts
+++ b/src/clients/soap/utils.ts
@@ -21,8 +21,9 @@ import os from "os";
 import path from "path";
 
 /**
+ * Optionally log out SOAP requests and responses for debugging purposes
  *
- * @param client
+ * @param client A SOAP client that generates requests and responses
  */
 const debugRequest = (client: Client) => {
   console.debug(client.lastRequest);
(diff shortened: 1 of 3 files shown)


src/clients/soap/utils.ts
```diff
@@ -20,27 +17,44 @@ import { createClientAsync, Client, SoapMethodAsync } from "soap";
 import axios from "axios";
 import { writeFile, rm } from "fs/promises";
 import { v4 as uuid } from "uuid";
+import os from "os";
+import path from "path";
 
+/**
+ *
+ * @param client
+ */
```
ryanwersal, line 26: This intended to be filled in or removed?

src/clients/soap/utils.ts
```diff
@@ -96,16 +124,19 @@ const overrideClientDefaults = (
   }
 };
 
+/**
+ * Make a request to a SOAP-based API
+ * @param param0 An object containing how to connect to a SOAP-based API, the method to onvoke, any overridden values for the HTTP client, and whether or not to log debug messages
```
ryanwersal, line 129: Nit: `onvoke` -> `invoke`

ryanwersal, line 129: Would it be better to move this part of the docs to the `RequestParams` type itself? I think that would provide better in-editor context though I'm unsure about the actual doc generation tooling.

ryanwersal (approved):
LGTM 👍

hipplej (approved):
LGTM

src/clients/soap/utils.ts
taylorreece: I think you're right - you can command-click the argument to get more detailed info about each destructured element.

PR #42, opened 2022-01-21
Add reasonable defaults for primitive fields

ryanwersal:
Conditional and Connection will require more elaborate defaulting - those
will work as before. The rest have reasonable defaults to be used in the
event the component author left it unspecified.


hipplej (approved):
LGTM

PR #43, opened 2022-02-11
Allow specifying a separate label for Key in keyvaluelist inputs

ryanwersal:
`label` can now accept either a `string`, as before, or a more complex
object of `{key: string; value: string}` to allow defining labels for
both labels in `keyvaluelist` collection inputs.

`key` is only passed on to the server if `collection` is set to `keyvaluelist`
though it will accept both forms - `key` is dropped at submission-time.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "5.3.0",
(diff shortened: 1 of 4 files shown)


hipplej (approved):
LGTM

PR #46, opened 2022-03-30
Add yeoman generator for Components

ryanwersal:
Includes sub-generators for individual connection, action, and trigger
files so we can scaffold out individual files after the initial Component
generation.


packages/generator-spectral/src/generators/connection/templates/oauth.ts
```diff
@@ -0,0 +1,65 @@
+import {
+	connection,
```
bkegley, line 2: formatting?

bkegley (approved):
Nice! LGTM!

PR #47, opened 2022-03-31
Fix handful of issues with new generator

ryanwersal:
Add missing template files, update templated configs, and make some
templates more real.


bkegley (approved):
LGTM

PR #48, opened 2022-04-05
Improve component definition type safety

ryanwersal:
- Make server types conversion more explicit (enabling better authoring
  types/DX)
- Update dependencies, particularly Typescript
- Add foundation for HTTP client with axios (support debug/retry)
- Add ability to define per-input clean function to move that logic from
  each action into the input definition; also types the perform param
  according to the return type of the clean function
- Add component-level error handler to move general error handling from
  each action into the component definition
- Refactor many multi-generic types to use `this` references and avoid
  the need for further generics
- Add `serializeObject` helper function to help deal with recursive data
  structures and various error handling scenarios


packages/spectral/src/util.ts
```diff
@@ -365,6 +370,14 @@ export const lowerCaseHeaders = (
...
   }, {});
 
+/** This function accepts an arbitrary object/value and safely serializes it.
+ *
+ * @param value Arbitrary object/value to serialize.
+ * @returns JSON serialized text that can be safely logged.
+ */
+export const serializeObject = (value: unknown): string =>
+  JSON.stringify(serializeError(value), null, 2);
+
```
hipplej, line 380: I am confused about the purpose of this function. Its name does not seem to match up with its implementation.

hipplej (approved):
Mostly just minor questions. Looks like really good improvements overall!

packages/spectral/src/util.ts
ryanwersal: Definitely open to suggestions - the intent here is to provide utilities to aid in handling arbitrary payloads, particularly for serialization use cases. We've seen several usages of `JSON.parse(JSON.stringify(...))` in an effort to "strip out" unserializable values (functions, cyclic references, and so on). `serializeError` handles the cycles and a few other use cases though it does have some behaviors that are likely undesirable for this use case. 

I'll look at finding a more suitable replacement.

packages/spectral/src/clients/http/inputs.ts
```diff
@@ -0,0 +1,169 @@
...
+export const headers = input({
+  label: "Header",
+  placeholder: "Header",
+  type: "string",
+  collection: "keyvaluelist",
+  required: false,
+  comments: "A list of headers to send with the request.",
+  example: "User-Agent: curl/7.64.1",
+  // clean: (value) =>
+  //   util.types.keyValPairListToObject(value, util.types.toString),
```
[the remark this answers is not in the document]
ryanwersal: I'll remove the `clean` function from this PR until we can figure out the type narrowing issues.

PR #56, opened 2022-04-20
Add clean function to inputs

ryanwersal:
This should allow moving most type coercion out of action perform functions
and into the input definitions themselves allowing for actions to focus on
their logic rather than duplicating arbitrary input type handling.

This also incorporates a refactoring to move as much input types knowledge
into the input interfaces themselves to reduce the number of areas we have
to keep in sync when adding more input types and/or functionality.


hipplej (approved):
LGTM. :shipit:

PR #57, opened 2022-04-21
Restore some generics for Action/Trigger definition

ryanwersal:
It appears that the `this` references were resolved at a time such that
it couldn't correctly narrow to the specific return type of the trigger/
action and caused issues with `invoke`/`invokeTrigger` based tests.


taylorreece (approved):
LGTM; thanks! I'm excited for all these DX improvements!

PR #67, opened 2022-07-05
Add params to HTTP createClient

taylorreece:
Some APIs pass in things (like API key) via `params`, so it's useful to be able to set a generic `params` value for all HTTP requests. This change lets you pass in a `params` object into Spectral's HTTP client's `createClient` function, and those params will be used for all HTTP requests.


bkegley (approved):
LGTM

PR #85, opened 2022-08-17
Default inputs in test harness

ryanwersal:
Supports both defaulted values and running unspecified values through
clean functions, if defined. Mimics in-app defaulting behavior as
closely as possible.


hipplej (approved):
Thanks!

PR #86, opened 2022-08-18
Add context to Data Sources

hipplej:


ryanwersal (approved):
LGTM 👍

Only thought is the handling of the clean functions in `createPerform` since that relies on arg length but I _think_ we're fine since both the arity and (roughly) the types match for data sources and actions.

hipplej:
> LGTM 👍
> 
> Only thought is the handling of the clean functions in `createPerform` since that relies on arg length but I _think_ we're fine since both the arity and (roughly) the types match for data sources and actions.

Best I can tell this is handled correctly.

PR #87, opened 2022-08-19
Prefer operationId but use verb/path for action key/name generation

ryanwersal:


packages/generator-spectral/src/generators/formats/readers/openapi/actions.ts
```diff
@@ -78,9 +78,7 @@ const buildAction = (
     | OpenAPIV3_1.ParameterObject
   )[] = []
 ): Action => {
-  if (!operation.operationId) {
-    throw new Error(`Failed to find operationId for ${path} ${verb}`);
-  }
+  const operationName = camelCase(operation.operationId || `${verb} ${path}`);
```
taylorreece, line 81: Can the `key` have a space like that? I don't recall if we use that `operationName` for variable names.

taylorreece (approved):

packages/generator-spectral/src/generators/formats/readers/openapi/actions.ts
ryanwersal: It all gets run through `camelCase` at the end so a `POST /foo/bar/baz/{id}` becomes `postFooBarBazId` - not great but it should at least successfully generate the component.

taylorreece: Ah; perfect!

PR #89, opened 2022-08-22
OpenAPI generator enhancements

ryanwersal:
- Wrap all input data objects in the `input` function to aid in perform fn typing
- Copy in the OpenAPI spec used for component generation
- Fix number/boolean inputs incorrectly using undefined for falsey values
- Escape double quotes in input comments
- Generated components now have a proper jest config
- Generated components also have a minimal start of a harness-based testing file templated out


taylorreece (approved):
Love it!

PR #95, opened 2022-09-27
Relax restrictions for raw request data payloads

ryanwersal:


taylorreece (approved):
LGTM; this'll allow users to reference an object from a previous step.

PR #96, opened 2022-09-29
Remove util.docs.formatJsonExample

taylorreece:
`util.docs.formatJsonExample` was only really useful for internal Prismatic component documentation for built-in components, and we've moved all of those to use `type: "code", language: "json"` for proper syntax highlighting. I think we can safely remove this utility function.


hipplej (approved):
Works for me. I'll be bumping the Spectral version momentarily anyway, so no worries about that.

PR #98, opened 2022-10-17
Add toObject util function to easily turn JSON into a parsed object

taylorreece:
When an input is a complex data structure, sometimes you don't know if the input you receive will be JSON-serialized or not.

This new utility function, `util.types.toObject()`, detects and parses JSON if it is passed JSON. Else, it returns the object that it received. That way, component users can input static JSON-serialized objects, or they can reference JavaScript objects from previous steps.

### DIFF
--- a/packages/generator-spectral/package.json
+++ b/packages/generator-spectral/package.json
@@ -37,7 +37,7 @@
     "generators"
   ],
   "dependencies": {
-    "@prismatic-io/spectral": "7.2.0",
+    "@prismatic-io/spectral": "7.2.1",
     "yeoman-generator": "5.6.1",
     "lodash": "4.17.21",
     "@apidevtools/swagger-parser": "10.1.0",
--- a/packages/spectral/package.json
+++ b/packages/spectral/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "7.2.0",
+  "version": "7.2.1",
   "description": "Utility library for building Prismatic components",
   "keywords": [
     "prismatic"
--- a/packages/spectral/src/util.test.ts
+++ b/packages/spectral/src/util.test.ts
@@ -871,4 +871,16 @@ describe("util", () => {
       expect(util.types.isConnection(v3)).toStrictEqual(false);
     });
   });
+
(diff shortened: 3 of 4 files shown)


hipplej (approved):

coltonmccormack:
@taylorreece Thoughts on how to approach types like null, arrays, and non-object valid JSON like `["foo","bar"]` with this?

We've been using the following internally for a few weeks, but the term "object" is admittedly a bit nebulous. In our case, we were always expecting a type matching `Record<string, any>`/`Object` rather than just `object` so that we could trust operations like `const foo = input_val["id"]` and not have to worry about having cases for undefined, null, arrays, functions, etc. that JS counts as an `object` in our components. Checking for `Object` types in JS is a strange rabbit hole.

```typescript
function toObject(obj: unknown): Object {
  if (typeof obj == "string" && util.types.isJSON(obj)) {
    obj = JSON.parse(obj);
  }

  if (
    obj !== null &&
    obj !== undefined &&
    Object.prototype.toString.call(obj).slice(8, -1) === "Object"
  ) {
    return obj as Object;
  } else {
    throw new Error(`Value '${JSON.stringify(obj)}' cannot be coerced to Object.`);
  }
}
```

taylorreece:
That's a great observation @coltonmccormack . "Object" is definitely a nebulous term. The goal with this utility function was to take inputs that are possibly JSON-serialized and deserialize them if they are JSON, but otherwise leave the input alone. Our [collection tools](https://prismatic.io/docs/components/collection-tools/) component, for example, has several inputs where we expect an array or `Record<string, any>`, but we wanted to support static string inputs in addition to references to previous steps' results. This function should deserialize arrays as well as `Record<string, any>`.

I think the function you pasted in seems reasonable. If an object is null or undefined, throw an error. For your inputs, do they have a well-defined format (like, keys you expect, etc?). You can leverage a `clean` function to check for certain object formats:
https://github.com/prismatic-io/examples/blob/main/components/data-example/src/index.ts#L10-L26
Some of our users have reached for object validation packages like [joi](https://joi.dev/api/?v=17.6.1) to validate that the object their input received was correct, and they throw an error if not.

PR #106, opened 2022-12-28
Strongly type JSON Forms schema and UI schema

taylorreece:
This provides better typing for JSON Forms schema and UI schema, to provide component developers more direction on what fields are needed. It also removes an unnecessary `data` property.


hipplej (approved):

ryanwersal (approved):
Thanks!

PR #110, opened 2023-02-09
Fix blank data payloads for raw request actions

taylorreece:
When a raw request action makes a request to an HTTP endpoint, but does not have a payload, `payload` is an empty string, `""`, which axios apparently serializes into `"\"\""`. That's usually fine -- you either specify a real payload, or you are doing a GET or DELETE request, and the HTTP body is ignored by the server. 

_However_, some APIs (looking at you, Shopify) will reject a GET or DELETE request if it contains any body in the request.

If the payload is an empty string, this causes the request to actually send nothing (and not the two-character `""` as the body).

### DIFF
--- a/packages/spectral/package.json
+++ b/packages/spectral/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "7.3.6",
+  "version": "7.3.7",
   "description": "Utility library for building Prismatic components",
   "keywords": [
     "prismatic"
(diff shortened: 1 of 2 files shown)


ryanwersal (approved):
LGTM; you'll have to relock the package since the version number gets encoded into it.

taylorreece:
> you'll have to relock the package since the version number gets encoded into it

We use yarn for spectral, so I think we're good here - no npm package lock.

ryanwersal:
> > you'll have to relock the package since the version number gets encoded into it
> 
> We use yarn for spectral, so I think we're good here - no npm package lock.

Ah - you're very right! :shipit:

PR #114, opened 2023-03-20
Add trigger/action attributes for User Level Config

hipplej:


taylorreece (approved):
Wonderful; thanks!

packages/spectral/src/types/ActionPerformFunction.ts
```diff
@@ -43,4 +44,6 @@ export interface ActionContext {
   customer: Customer;
   /** Contains attributes of the Instance that is being executed. */
   instance: Instance;
+  /** Contains attributes of the User for whom a User Level Configuration is being used. */
+  user: User;
```
taylorreece, line 48: Will user always be set, or just for ULC? Should this be `user?: User`?

hipplej: I guess the way I envisioned it is that it would always be there but the properties contained within would be `null`. Seems like that's easier to work with than having to check for existence as well?

PR #123, opened 2023-04-13
Add an input for form data file names

taylorreece:
When you include a file in a form-data POST or PUT request, that file can include a `filename`. That `filename` is often used by a third-party API to determine MIME type, present the file in their app, etc.

Up until now, the `filename` was always set to the file's `key`. That resulted, in many apps, in a file simply named `file`.

This allows you to optionally specify a name for a file that you're uploading.

Screen Shot 2023-04-13 at 10 32 26 AM

### DIFF
--- a/packages/spectral/package.json
+++ b/packages/spectral/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "7.6.0",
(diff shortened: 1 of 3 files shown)


ryanwersal (approved):
LGTM 👍

PR #129, opened 2023-05-23
Add Unix epoch support for util.types.toDate() and support string dates for comparison operators

taylorreece:
The library that we use for datatime comparison, `date-fns`, dropped support for string dates (like `2023-05-23`) in v2.0.0. This PR expands our support for various timestamp formats. We can now compare `Date()` objects, ISO strings (like `"2030-03-20T12:50:30.105Z"`), or Unix epoch timestamps (like `1631568050` or its string equivalent).

### DIFF
--- a/packages/spectral/package.json
+++ b/packages/spectral/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "7.6.5",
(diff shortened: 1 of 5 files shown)


ryanwersal (approved):
LGTM 👍

PR #130, opened 2023-06-05
Handle operationIds with leading numbers

ryanwersal:
Variables cannot start with numbers so convert those to text, similar to what was done for the "action groups" (filenames).

### DIFF
--- a/packages/generator-spectral/src/generators/formats/readers/openapi/actions.ts
+++ b/packages/generator-spectral/src/generators/formats/readers/openapi/actions.ts
@@ -1,9 +1,9 @@
 import { OpenAPIV3, OpenAPIV3_1 } from "openapi-types";
 import { camelCase, isEmpty, startCase } from "lodash";
-import { toWords } from "number-to-words";
 import { getInputs } from "./inputs";
 import { Action, Input, stripUndefined } from "../../utils";
 import { WriterFunction } from "ts-morph";
+import { toGroupTag, toOperationName } from "./util";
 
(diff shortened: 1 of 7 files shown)


packages/generator-spectral/src/generators/formats/readers/openapi/actions.ts
```diff
@@ -68,6 +68,14 @@ const toGroupTag = (path: string): string =>
     (_, match) => toWords(match)
   );
 
+const toOperationName = (operationName: string): string =>
+  camelCase(
+    operationName.replace(
+      /^([0-9]+)(.+)$/,
+      (_, num, rest) => `${toWords(num)}${rest}`
```
taylorreece, line 75: Will this potentially generate variable names with commas and spaces? 

```
// Large numbers:
converter.toWords(9007199254740992); // => “nine quadrillion, seven trillion, one hundred ninety-nine billion, two hundred fifty-four million, seven hundred forty thousand, nine hundred ninety-two”
```

taylorreece: Maybe sufficient to wordify the leading number? Just remove the `+` in the regex?

ryanwersal: `camelCase` handles normalizing a lot of these cases but I can see where converting _all_ numbers is overkill. I'll reduce it to the single digit.

taylorreece (approved):
Latest LGTM!

PR #131, opened 2023-06-14
isBufferDataPayload should return true only if the data property is a Buffer

taylorreece:
`isBufferDataPayload` was misidentifying objects of the form 
```
{
  "data" : {
    "foo": "bar"
  }
}
```
as `DataPayload` objects because they are an object with a `data` property.

But, the check is missing a critical piece: the `data` property needs to be a `Buffer`. This change checks that the object's `data` property is a `Buffer`.

### DIFF
--- a/packages/spectral/package.json
+++ b/packages/spectral/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "7.6.6",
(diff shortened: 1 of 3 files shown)


bkegley (approved):

ryanwersal (approved):
LGTM 👍

PR #136, opened 2023-07-27
Ensure correct docs links for public components

taylorreece:
Mostly an internal-focused update. All of our public components have documentation links of the form `https://prismatic.io/docs/components/KEY`. We should put up guard rails to ensure new public components all have correct documentation URLs.


ryanwersal (approved):
LGTM 👍

PR #137, opened 2023-08-11
Evaluate conditional predicates from left to right like JavaScript, rather than all at once

taylorreece:
Suppose you have an integration that might receive an array as part of its body. You want to create branching logic to say: `if this array is included in the payload, and if "foo" is included in the array, follow this branch`. You could add branching logic that looks like this:

image

The JavaScript equivalent to this branch logic would be something like `if ( payload?.myArray && myArray.includes("foo") ) { doThing(); }`

That doesn't currently work with the way we process predicates. We currently evaluate all predicates (the "if this array exists" and "if this array contains 'foo'") at once, and then `reduce` the result. In this case, if an array doesn't exist the branch step will throw an error about the non-existent array not being an array - `Execution of 'Branch on Expression' failed with error: Incompatible comparison arguments, Error: Invalid arguments set to 'contains'.`

This changes aligns our `and` and `or` conditional logic more closely with a programming language. It will loop over predicates. In the case of `and`, if a `false` is encountered, it'll return `false` (otherwise, `true`). Similarly, in the case of `or`, if a `true` is encountered it'll return `true` (otherwise, `false`). 

The logic doesn't change from what we had before, but we stop processing predicates once we know that we can return `true` or `false`, like JavaScript and other languages do.

Now, with an `and` after checking if the array exists, and finding it doesn't, it'll return `false` without trying to determine if a  value is contained in a non-existent array.

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "all": "turbo run all"
   },
   "devDependencies": {
-    "turbo": "^1.8.5"
+    "turbo": "^1.10.12"
   },
   "resolutions": {
     "**/eslint-plugin-import/tsconfig-paths/json5": "2.2.3",
--- a/packages/spectral/package.json
+++ b/packages/spectral/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "7.6.9",
+  "version": "7.7.0",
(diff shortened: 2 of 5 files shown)


ryanwersal (approved):
LGTM 👍

adriancarriger (approved):
✅

PR #145, opened 2023-09-15
Add a few new attributes to action contexts and trigger payloads

hipplej:


taylorreece:
For getting a trigger's own webhook URL, they can call `context.webhookUrls[context.flow.name]`, then, right?

taylorreece (approved):

PR #151, opened 2023-11-02
Do not template inputs for readOnly properties in OpenAPI-generate components

taylorreece:
An object can be defined in an OpenAPI spec with properties that are `readOnly`. It's handy to be able to reference that same object for a `requestBody` and for a return value (to prevent having to maintain two versions of the same object). But, we're currently templating out inputs for `readOnly` properties. We should filter out the `readOnly` properties as we generate inputs.


ryanwersal (approved):
LGTM 👍

PR #167, opened 2023-12-13
More strictly define collection/default type combinations for different InputFields

robertmaloney:
This PR focuses on two goals:
- The type of `default` is now known statically for all InputFields.
- The types of the `collection` and `default` properties are correctly coupled.

Other notes:
- `DataInputField` uses `Buffer` as its generic type instead of `string` like the other inputs.
- The `clean` functions remain the type `InputCleanFunction<unknown(, unknown)>`.

### DIFF
--- a/packages/spectral/src/types/Inputs.ts
+++ b/packages/spectral/src/types/Inputs.ts
@@ -109,102 +109,110 @@ interface BaseInputField {
   required?: boolean;
 }
 
-export interface StringInputField extends BaseInputField {
-  /** Data type the InputField will collect. */
-  type: "string";
+type CollectionOptions<T> =
+  | SingleValue<T>
+  | ValueListCollection<T>
+  | KeyValueListCollection<T>;
+
+interface SingleValue<T> {
   /** Collection type of the InputField */
-  collection?: InputFieldCollection;
+  collection?: undefined;
   /** Default value for this field. */
-  default?: unknown;
-  /** Dictates possible choices for the input. */
-  model?: InputFieldChoice[];
-  /** Clean function */
-  clean?: InputCleanFunction<this["default"]>;
+  default?: T;
 }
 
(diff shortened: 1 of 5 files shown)


packages/spectral/src/types/Inputs.ts
```diff
@@ -109,102 +109,110 @@ interface BaseInputField {
...
+    /** Data type the InputField will collect. */
+    type: "string";
+    /** Dictates possible choices for the input. */
+    model?: InputFieldChoice[];
+    /** Clean function */
+    clean?: InputCleanFunction<unknown>;
+  };
+
+export type DataInputField = BaseInputField &
+  CollectionOptions<Buffer> & {
```
ryanwersal, line 149: `data` is more often used for indicating larger data payloads or data passed by reference. Defaults are also delivered as strings without further massaging of the data so I don't think we can do anything other than `string` here as well, at least without platform changes.

packages/spectral/src/types/Inputs.ts
```diff
@@ -109,102 +109,110 @@ interface BaseInputField {
...
   /** Collection type of the InputField */
-  collection?: InputFieldCollection;
+  collection: "keyvaluelist";
   /** Default value for this field. */
-  default?: unknown;
-  /** Dictates possible choices for the input. */
-  model?: InputFieldChoice[];
-  /** Clean function */
-  clean?: InputCleanFunction<this["default"]>;
+  default?: { key: string; value: T }[];
```
ryanwersal, line 135: It would probably be most proper to use the `interface KeyValuePair<V = unknown>` type here.

ryanwersal (approved):
Looks good! Suggested a couple small tweaks. This PR should also bump the package version and relock the lockfile as we'll ship this soon as it lands.

PR #174, opened 2024-01-31
Rework config vars/pages to streamline creation/usage for CNI

ryanwersal:
Remove configVars in favor of defining inline with configPages. Update types to flow through the entire ConfigPages collection which enables richer typing for testing flows and within onTrigger/onExecution functions.


hipplej (approved):
LGTM!

PR #175, opened 2024-02-01
Fix and improve types for configVars and CNI conversions

ryanwersal:


packages/spectral/src/serverTypes/convertIntegration.ts
```diff
@@ -343,8 +351,10 @@ const codeNativeIntegrationComponent = (
               "The function that will be executed by the flow to return an HTTP response.",
           },
           perform: onTrigger,
-          onInstanceDeploy: onInstanceDeploy,
-          onInstanceDelete: onInstanceDelete,
+          onInstanceDeploy,
+          hasOnInstanceDeploy: !!onInstanceDeploy,
+          onInstanceDelete,
+          hasOnInstanceDelete: !!onInstanceDelete,
```
robertmaloney, line 357: When are these new `has*` properties used?

robertmaloney (approved):
Just had a question, I don't know how valuable my approval is here given my lack of broader context but it LGTM.

packages/spectral/src/serverTypes/convertIntegration.ts
ryanwersal: They indicate the existence of these deploy/delete functions to the API so we can optimize later checks without having to consult the actual component code.

PR #176, opened 2024-02-06
On Premise Connection Configuration

robertmaloney:
Adds `onPremiseConfig` optional metadata to enable on prem variants of connections. One shortcoming of the currently implementation is that the `replacesInputs` type doesn't have enough type inference to restrict it's array members to the keys of the `inputs` field. An attempt to address that exists on branch `rm/sc-18076/onprem-connections-generics` but it has tradeoffs.

### DIFF
--- a/packages/generator-spectral/package.json
+++ b/packages/generator-spectral/package.json
@@ -40,7 +40,7 @@
   ],
   "dependencies": {
     "@apidevtools/swagger-parser": "10.1.0",
-    "@prismatic-io/spectral": "8.0.3",
+    "@prismatic-io/spectral": "8.0.4",
     "lodash": "4.17.21",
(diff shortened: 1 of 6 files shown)


taylorreece (changes requested):
Thanks @robertmaloney ! Could you please bump `packages/spectral/package.json` version to 8.0.4 and run `yarn` from the base directory to bump the spectral package version?

ryanwersal:
This has been updated with the reworked on-premise available/controlled implementation we discussed. It also addresses @taylorreece 's request for bumping the version and relocking.

taylorreece (approved):

robertmaloney:
Can't formally approve but LGTM

PR #187, opened 2024-03-18
Simplify http debug logging

taylorreece:
Currently, if you select `debug: true` in a `createClient` function, you can run out of memory if you request sufficiently large binary files (i.e. a several MB PDF file). This is due to the JSON serialization of the PDF file's `Buffer`.

This PR simplifies or extends a few things:

1. When `debug: true` is enabled for an HTTP client, extraneous properties of `request` and `response` are no longer logged. Axios provides lots of metadata about the request (like `maxFreeSockets` that simply aren't helpful when debugging a request in Prismatic. Relevant properties, like base URL, url, HTTP method, headers, status code and payload are retained. 
2. When a payload is "large" (> 10KB) or when it's a binary file (represented as a `Buffer`), the data will be represented as `<data (10000000 bytes)>`. Headers, URL, etc., will still be accessible.
3. Key order will be retained when logging out request/response (so, you'll see the most important things like URL and headers first, data after). `toJSON` has been updated so that you can optionally retain key order.

### DIFF
--- a/packages/spectral/package.json
+++ b/packages/spectral/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "8.0.7-rc.2",
+  "version": "8.0.7",
   "description": "Utility library for building Prismatic components",
   "keywords": [
     "prismatic"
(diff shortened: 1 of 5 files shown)


ryanwersal (approved):
LGTM, thanks!

PR #196, opened 2024-04-26
Fix the contains function to handle stringified numbers

taylorreece:
The `contains` function is designed to determine if an element is contained within an array (or object, or string... but this PR relates to arrays).

For strings, it works as expected - `"foo"` is contained in `["foo", "bar"]`.
For numbers, it works as expected - `2` is contained in `[1,2,3]`.
For stringified numbers, it does not work as expected. It currently returns `false` if you check if `"2"` is contained in `["1", "2", "3"]`.

That's because the element you check is passed through the `parseValue` function, which checks if the value is a string and, if so, the value is `JSON.parse`'d. `JSON.parse("2")` becomes `2`, and `2 !== "2"`.

This PR loosens restrictions on contains, so that the value `"2"` would be passed to `contains`, and both the values `"2"` and `2` would be searched for in the array.

### DIFF
--- a/packages/spectral/package.json
+++ b/packages/spectral/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "8.1.3",
+  "version": "8.1.4",
   "description": "Utility library for building Prismatic components",
   "keywords": [
     "prismatic"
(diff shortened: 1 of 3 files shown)


ryanwersal (approved):
LGTM 👍

PR #200, opened 2024-05-24
Refine collection type handling for CNI config vars

ryanwersal:
This better aligns CNI `collectionType` configuration with the Designer's; no `collectionType` for schedule or connection `dataType`s or for a few types of data sources.

### DIFF
--- a/packages/spectral/src/types/IntegrationDefinition.ts
+++ b/packages/spectral/src/types/IntegrationDefinition.ts
@@ -15,6 +15,7 @@ import {
   ObjectSelection,
   ConfigVarResultCollection,
   Schedule,
+  CollectionDataSourceType,
 } from ".";
 import { Prettify, UnionToIntersection, ValueOf } from "./utils";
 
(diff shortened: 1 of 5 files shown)


packages/spectral/src/types/IntegrationDefinition.ts
```diff
@@ -217,10 +237,14 @@ export type ConnectionConfigVar<TComponents extends ComponentSelector<any>> =
 
 export type ConfigVar<TComponents extends ComponentSelector<any>> =
   | StandardConfigVar
+  | CodeConfigVar
   | ScheduleConfigVar
   | DataSourceConfigVar<TComponents>
   | ConnectionConfigVar<TComponents>;
 
+export const isCodeConfigVar = (cv: ConfigVar<any>): cv is CodeConfigVar =>
```
robertmaloney, line 245: Where is this going to be used?

robertmaloney (approved):

packages/spectral/src/types/IntegrationDefinition.ts
ryanwersal: I was initially using it in the `convertIntegration` layer but that ended up not being strictly necessary. I opted to leave it since it's now a formal config var type and the rest of that union have type guard functions.

PR #201, opened 2024-05-24
Fix handling of collection type connection inputs

ryanwersal:


robertmaloney (approved):

packages/spectral/src/serverTypes/convertIntegration.ts
```diff
@@ -184,7 +184,16 @@ const convertComponentReference = <TValue>({
 
   const inputs = Object.entries(values ?? {}).reduce((result, [key, value]) => {
     if ("value" in value) {
-      return { ...result, [key]: { type: "value", value: value.value } };
+      const type = value.value instanceof Object ? "complex" : "value";
+      const valueExpr =
+        value.value instanceof Object && !Array.isArray(value.value)
+          ? Object.entries(value.value).map<ServerInput>(([key, value]) => ({
```
robertmaloney, line 190: `key` and `value` are already in the current scope due to the top-level `reduce`, are we comfortable with shadowing here?

packages/spectral/src/serverTypes/integration.ts
```diff
@@ -12,13 +12,13 @@ export interface ComponentReference {
 
 export type Input =
   | {
-      name?: string;
+      name?: string | Input;
```
robertmaloney, line 15: This could allow an infinite nesting within `name` right? Should we hoist `name`'s type up so we can enforce only one level?

ryanwersal: It would. It's used as a guide for the conversion layer so I don't think avoiding the possible recursion is necessary (since it isn't on the DX "authoring" side).

PR #202, opened 2024-06-13
Add a Component Manifest generator

jasoncomes:
Addition of a Component Manifest generator to the Spectral tool, which is used for the integration between the Component and Code Native Integration (CNI) processes.

The key features of this addition include:

1. A command-line interface (CLI) executable with the following flags:
   - `--help`: Provides information about the available commands and options.
   - `--dry-run`: Allows for a test run without actually generating the manifest.
   - `--signature`: Generates a signature for the manifest.

2. The ability to run the CLI executable from within the `Component` directory where Spectral is installed. This allows the tool to retrieve the necessary data from the `Component` entry.

3. A scaffolding data and EJS (Embedded JavaScript) template generator, which includes the following templates:
   - EJS Static Templates
   - EJS Action Templates
   - EJS Datasource Templates
   - EJS Connection Templates
   - EJS Trigger Templates

These templates are used to generate the Component Manifest, which is a crucial part of the integration between the Component and Code Native Integration (CNI) processes.

Includes [Code Scaffolding](https://app.shortcut.com/prismatic/story/21306/enable-code-scaffolding-from-spectral)


packages/spectral/package.json
```diff
@@ -23,6 +26,9 @@
   "scripts": {
     "clean": "rm -rf dist",
     "build": "yarn run format && yarn run lint && yarn run clean && tsc",
+    "build:templates": "copyfiles --error -u 1 src/**/templates/**/* dist",
+    "cli:permission": "chmod +x dist/generators/**/cli.js",
```
jasoncomes, line 30: When running the command `npm run build` to generate the Spectral package, the file permissions are being stripped as the files are added to the `dist` directory. We need to ensure that the command-line interface (CLI) executables in the generators directory have the correct executable permissions. While the `chmod` command works in Linux environments, we're not sure if it's compatible with our Continuous Integration (CI) environments. Therefore, we'll need to find a more suitable command to set the necessary permissions for those CLI executables.

PR #205, opened 2024-06-18
Add additional flags, new flag functionality, component verification, and cleanup

jasoncomes:
Add additional flags, new flag functionality, component verification, and cleanup

- Verify component object before creating manifest
- Cleanup existing manifest if exists
- Add --name flag
- Add --output-dir flag
- Add flag aliases
- Consolidate Flags
- Clean up unused code


packages/spectral/src/generators/componentManifest/constants.ts
```diff
@@ -1,36 +1,54 @@
...
-export const DESTINATION_BASE_NAME = `${path.basename(COMPONENT_DIR)}-manifest`;
-export const DESTINATION_DIR = path.join(
-  COMPONENT_DIR,
-  "..",
-  DESTINATION_BASE_NAME
-);
+
+export const DESTINATION_BASE_NAME = FLAGS.NAME.value
+  ? `${FLAGS.NAME.value}-manifest`
+  : `${path.basename(COMPONENT_DIR)}-manifest`;
```
bkegley, line 47: We should only append `-manifest` if there isn't a provided name. Otherwise, we should just use the name.

bkegley (approved):

PR #219, opened 2024-06-26
CNI - Visibility and Permissions for Config Vars & Inputs

jasoncomes:
Add visibility and permissions feature for Config Vars & Inputs.

### DIFF
--- a/packages/spectral/src/serverTypes/convertIntegration.ts
+++ b/packages/spectral/src/serverTypes/convertIntegration.ts
@@ -17,6 +17,7 @@ import {
   ComponentRegistry,
   ComponentManifest,
   ComponentManifestAction,
+  PermissionAndVisibilityType,
 } from "../types";
 import {
   Component as ServerComponent,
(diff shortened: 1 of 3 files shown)


packages/spectral/src/serverTypes/convertIntegration.ts
```diff
@@ -170,6 +180,85 @@ const codeNativeIntegrationYaml = (
...
+const getPermissionAndVisibilityValues = ({
+  permissionAndVisibilityType = "customer",
+  visibleToOrgDeployer = true,
+}: {
+  permissionAndVisibilityType?: PermissionAndVisibilityType;
+  visibleToOrgDeployer?: boolean;
+}) => {
+  return {
+    ...permissionAndVisibilityTypeValueMap[permissionAndVisibilityType],
+    ...(visibleToOrgDeployer !== null ? { visibleToOrgDeployer } : {}),
```
bkegley, line 219: This should be checking for `undefined` instead of `null`

packages/spectral/src/serverTypes/convertIntegration.ts
```diff
@@ -170,6 +180,85 @@ const codeNativeIntegrationYaml = (
   return YAML.stringify(result);
 };
 
+type VisibilityAndPermissionValue = {
+  [Key in `visibleTo${Capitalize<PersonaType>}Deployer`]: boolean;
```
bkegley, line 184: This is fun but probably overly clever since it's the only place we're doing this, the only place we use `PersonaType`, and requires hopping back and forth between two files to figure out what it is.

bkegley (approved):

PR #220, opened 2024-06-26
Update KeyValueList Conversion, Updates Perform Types, and Updates Templates to Account for the Former

awburgard:


packages/spectral/src/generators/componentManifest/getInputs.ts
```diff
@@ -90,7 +92,29 @@ export const INPUT_TYPE_MAP: Record<InputFieldDefinition["type"], InputType> = {
...
+const isKeyValueList = (
+  input: ServerTypeInput
+): input is ServerTypeInput & { collection: "keyvaluelist" } => {
+  return input.collection === "keyvaluelist";
+};
+
+const isValueList = (
+  input: ServerTypeInput
+): input is ServerTypeInput & { collection: "valuelist" } => {
+  return input.collection === "valuelist";
```
bkegley, line 119: This is fine to do but we don't really need a type guard here as we're just checking the value of the optional `collection` property. Could change the above `if` statements to the following and remove these functions:

```ts
if (input.collection === "keyvaluelist") {
  return `Record<string, ${valueType}>`;
}
```

packages/spectral/src/generators/componentManifest/templates/partials/inputs.ejs
```diff
@@ -0,0 +1,19 @@
...
+    <%_ } else { -%>
+    collection: null,
+    <%_ } -%>
+    <%_ if (input.default) { -%>
+    default: <%= input.default %>,
+    <%_ } -%>
+    <%_ if (input.required) { -%>
+    required: <%= input.required %>,
+    <%_ } -%>
+    <%_ if (input.onPremiseControlled) { -%>
```
bkegley, line 15: I'm guessing this would omit all falsy values here when if `input.onPremiseControlled === false` we'd want to include that property with a `false` value.

packages/spectral/src/generators/componentManifest/templates/partials/inputs.ejs
```diff
@@ -0,0 +1,19 @@
+<% inputs.forEach((input) => { -%>
+  <%= input.key %>: {
+    inputType: "<%= input.defaultType %>",
```
bkegley, line 3: I'd rename `input.defaultType` to either `input.type` or `input.inputType`. `defaultType` communicates that the type could change.

bkegley (approved):

PR #227, opened 2024-06-27
CNI - Skip signature verification & update package name default

jasoncomes:
- Add flag `--skip-signature-verify` to allow uses to always create a signature in the component manifest, whether the component has been published or not to the platform.  
- Update the package name default to `@component-manifests/[component-key]` instead of `@components/[component-key]-manifest`

### DIFF
--- a/packages/spectral/src/generators/componentManifest/cli.ts
+++ b/packages/spectral/src/generators/componentManifest/cli.ts
@@ -40,11 +40,12 @@ export const runMain = async (process: NodeJS.Process) => {
     process.exit(1);
   }
 
+  const args = process.argv.slice(3);
   const flags = {
     name: {
       flag: ["--name", "-n"],
       value: getFlagsStringValue({
-        args: process.argv.slice(3),
+        args,
         flags: ["--name", "-n"],
       }),
       description:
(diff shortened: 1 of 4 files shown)


7AXEL:
Check out this project: https://github.com/7AXEL/Web2APK

packages/spectral/src/generators/componentManifest/cli.ts
```diff
@@ -122,8 +124,8 @@ export const runMain = async (process: NodeJS.Process) => {
   await createComponentManifest({
     component,
     dryRun: flags.dry_run.value,
-    includeSignature: flags.include_signature.value,
-    packageName: flags.name.value ?? `@components/${component.key}-manifest`,
+    validateSignature: flags.validate_signature.value,
+    packageName: flags.name.value ?? `@component-manifests/${component.key}`,
```
awburgard, line 128: i _think_ we'll need to update the Readme as well

jasoncomes: Nope, this `packageName` is passed down to the readme.

image

bkegley (approved):

PR #228, opened 2024-06-27
Add ULC support for CNI

bkegley:
This exposes another interface `IntegrationDefinitionUserLevelConfigPages` that expects to be augmented via the special root export `userLevelConfigPages`. Config variable types (in e.g. `ActionContext['configVars']`) do not widen to `{[key: string]: ConfigVar}` unless _both_ `configPages` and `userLevelConfigPages` aren't exported from the root.

### DIFF
--- a/packages/spectral/src/serverTypes/convertIntegration.ts
+++ b/packages/spectral/src/serverTypes/convertIntegration.ts
@@ -48,28 +48,35 @@ export const convertIntegration = (
   // inline as part of the integration definition.
   const referenceKey = uuid();
 
-  const configVars: Record<string, ConfigVar> = Object.values(
-    definition.configPages ?? {}
-  ).reduce<Record<string, ConfigVar>>(
-    (acc, configPage) =>
-      Object.entries(configPage.elements).reduce<Record<string, ConfigVar>>(
-        (acc, [key, element]) => {
-          // "string" elements are HTML elements and should be ignored.
-          if (typeof element === "string") {
-            return acc;
-          }
+  const configVars: Record<string, ConfigVar> = Object.values({
+    configPages: definition.configPages ?? {},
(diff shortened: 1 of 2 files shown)


packages/spectral/src/serverTypes/convertIntegration.ts
```diff
@@ -48,28 +48,35 @@ export const convertIntegration = (
...
+        (acc, configPage) =>
+          Object.entries(configPage.elements).reduce<Record<string, ConfigVar>>(
+            (acc, [key, element]) => {
+              // "string" elements are HTML elements and should be ignored.
+              if (typeof element === "string") {
+                return acc;
+              }
+
+              if (key in acc) {
+                throw new Error('Duplicate config var key "' + key + '"');
```
jasoncomes, line 67:
```suggestion
                throw new Error(`Duplicate config var key ${key}`);
``` 
This might be easier to read

packages/spectral/src/types/IntegrationDefinition.ts
```diff
@@ -50,6 +50,21 @@ export interface IntegrationDefinitionComponentRegistry {}
...
+ * The expected interface when augmenting is:
+ *
+ * ```ts
+ * interface IntegrationDefinitionUserLevelConfigPages {
+ *   [key: string]: ConfigPage
+ * }
+ * ```
+ *
+ */
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
```
jasoncomes, line 65: Wonder at this point if we just turn this rule off in the eslint rules `.eslintrc.json`

jasoncomes (approved):

packages/spectral/src/types/IntegrationDefinition.ts
bkegley: Nope. Pretty explicit that we only want to allow it here.

PR #237, opened 2024-07-02
removes dynmaicObjectSeletion and dyamicFieldSelection from input

awburgard:


bkegley (approved):

jasoncomes (changes requested):
Gonna block this for a second. Double checking something in here.

packages/spectral/src/generators/componentManifest/getInputs.ts
```diff
@@ -34,11 +34,10 @@ interface GetInputsProps {
 
 export const getInputs = ({ inputs, docBlock }: GetInputsProps): Input[] => {
   return inputs.reduce((acc, input) => {
-    const valueType = getInputValueType(input);
```
jasoncomes: The `getInputValueType` returns a value from this map using `input.type`, believe here we only need to check against that `input.type` whether to include it or not.

```
export const INPUT_TYPE_MAP: Record<InputFieldDefinition["type"], ValueType> = {
  string: "string",
  data: "string",
  text: "string",
  password: "string",
  boolean: "boolean",
  code: "string",
  conditional: {
    type: "ConditionalExpression",
    module: "@prismatic-io/spectral",
  },
  connection: {
    type: "Connection",
    module: "@prismatic-io/spectral",
  },
  objectSelection: {
    type: "ObjectSelection",
    module: "@prismatic-io/spectral",
  },
  objectFieldMap: {
    type: "ObjectFieldMap",
    module: "@prismatic-io/spectral",
  },
  jsonForm: {
    type: "JSONForm",
    module: "@prismatic-io/spectral",
  },
  dynamicObjectSelection: "string",
  dynamicFieldSelection: "string",
};
```

PR #242, opened 2024-07-15
Improve CNI data sources

bkegley:


packages/spectral/src/types/ComponentRegistry.ts
```diff
@@ -107,15 +107,22 @@ type ComponentReferencesByType = UnionToIntersection<
...
+                                      component: TComponentKey;
+                                      key: TComponentPropertyKey;
+                                      values: {
+                                        [Key in keyof TInputs]: ComponentReferenceTypeValueMap<
+                                          TInputs[Key]
+                                        >[TComponentReferenceType];
+                                      };
+                                    }>;
+                                  }
+                                >
```
bkegley, line 125: This is the bulk of the change. Instead of creating a union of references from all the manifest action/trigger/data sources/connections it includes the metadata provided in the manifest (e.g. `inputs`, `onPremAvailable`, `dataSourceType`, etc.). That gives us access to using these properties in type creation downstream.

awburgard (approved):

PR #245, opened 2024-07-15
CNI Action Calling

jasoncomes:
- Introduce a Component Manifest Generator, which creates component-specific wrapper interfaces that developers can utilize in Code Native Integrations.
- Implement a Component Registry interface to inherit the component-specific wrapper interfaces through module augmentation.
- Update the type interfaces for Config Variables.
- Modify the conversion and invocation layer for components and integrations to accommodate the new interfaces and data coming from the component manifests.


bkegley (approved):
:shipit:

PR #246, opened 2024-07-17
Bump TS to 5.5.3 and migrate to biome

bkegley:
There are a couple relevant changes that make this upgrade necessary:

- bug fixes with type inference using union literal types (5.1) A good majority of the TS changes that make all the type inference possible revolve around creating large union literal types from component manifests. Prior versions of TS struggled to correctly narrow these union types progressively as a user partially satisfies the interface.

- vastly improve performance for conditional and union literal types (5.1) https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-1.html#negative-case-checks-for-union-literals

- added `extends` constraints when using `infer` (4.7) Previous versions of TS required a second `infer` statement in order to constrain an `extends infer X` statement. We utilize this pattern extensively which resulted in types that were overly difficult to grok.


packages/spectral/src/types/ConfigVars.ts
```diff
@@ -194,62 +181,49 @@ export type StandardConfigVar =
...
-  ComponentRegistryDataSource extends infer TDataSourceReference
-    ? TDataSourceReference extends ComponentRegistryDataSource
-      ? Omit<
-          BaseDataSourceConfigVar<TDataSourceReference["dataSourceType"]>,
-          "dataSourceType"
-        > & {
-          dataSource: TDataSourceReference["reference"];
-        }
-      : never
+  ComponentRegistryDataSource extends infer TDataSourceReference extends ComponentRegistryDataSource
```
bkegley, line 219: This change alongside the bump past TS 5.1 is really the fix here. `MyType extends infer TMyType extends MyType` is narrower than `MyType extends infer TMyType ? TMyType extends MyType ? ...`

packages/spectral-test/src/ComponentReference.test-d.ts
```diff
@@ -13,7 +13,7 @@ type AllComponentReferences =
   | ActionReference;
 
 const createRef = <TComponentRef extends AllComponentReferences>(
-  ref: TComponentRef
+  ref: TComponentRef,
```
awburgard, line 16: huge nit here: is this from biome formatter? just seems odd to have a , here

packages/spectral-test/biome.jsonc
```diff
@@ -0,0 +1,38 @@
...
+    "enabled": true
+  },
+  "files": {
+    "ignore": ["dist", "node_modules"]
+  },
+  "formatter": {
+    "enabled": true,
+    "indentStyle": "space",
+    "indentWidth": 2,
+    "lineWidth": 100
```
jasoncomes, line 13: I see what you did here! I take 100 over 80 anyday.

packages/spectral-test/biome.jsonc
```diff
@@ -0,0 +1,38 @@
...
+        "useImportType": "off"
+      },
+      "complexity": {
+        "noForEach": "off"
+      },
+      "performance": {
+        "noAccumulatingSpread": "off"
+      }
+    }
+  }
```
jasoncomes, line 37:
```suggestion
  }
  "javascript": {
    "formatter": {
      "jsxQuoteStyle": "double",
      "quoteProperties": "asNeeded",
      "trailingComma": "all",
      "semicolons": "always",
      "arrowParentheses": "always",
      "bracketSpacing": true,
      "bracketSameLine": false,
      "quoteStyle": "double",
      "attributePosition": "auto"
    }
  }
```
You can additional formating options like so to better match our standard prettier setup.

packages/spectral-test/package.json
```diff
@@ -25,28 +23,19 @@
     "clean": "rm -rf dist",
     "build": "yarn run format && yarn run lint && yarn run clean && tsc",
     "prepack": "yarn run build",
-    "format": "yarn run lint-fix && prettier --loglevel error --write 'src/**/*.ts' '*.{ts,js,json}'",
-    "check-format": "prettier --check 'src/**/*.ts' '*.{ts,js,json}'",
-    "check": "yarn run check-format && yarn run lint",
-    "lint": "eslint --ext .ts .",
-    "lint-fix": "eslint --fix --ext .ts .",
+    "lint": "biome lint .",
+    "lint-fix": "biome lint --apply-unsafe .",
```
jasoncomes, line 27:
```suggestion
    "lint-fix": "biome lint --write --unsafe .",
```
`--apply-unsafe` is deprecated and recommended you use above.

packages/spectral-test/package.json
```diff
@@ -25,28 +23,19 @@
...
+    "format": "yarn run lint-fix && biome format --write .",
```
jasoncomes, line 28:
```suggestion
    "format": "biome check --write",
```
Will run both lint and format.

awburgard (approved):

packages/spectral/src/clients/http/index.ts
```diff
@@ -105,8 +101,8 @@ export const createClient = ({
     headers,
     timeout,
     params,
-    maxContentLength: Infinity,
-    maxBodyLength: Infinity,
+    maxContentLength: Number.POSITIVE_INFINITY,
+    maxBodyLength: Number.POSITIVE_INFINITY,
```
jasoncomes, line 105: lol namespaces, I like this. ^ @jpeters5392

packages/spectral/src/types/ConfigVars.ts
jasoncomes: Looks way better

packages/spectral-test/biome.jsonc
jasoncomes: Probably don't need these if they're already defaulted to these settings though in the global formatter property.

jasoncomes (approved):

packages/spectral-test/src/ComponentReference.test-d.ts
bkegley: Yup adds trailing commas

packages/spectral-test/biome.jsonc
bkegley: yagni :D

packages/spectral-test/package.json
bkegley: It's deprecated in future versions. Going to keep this consistent with all our other biome configurations.

PR #252, opened 2024-07-19
CNI - Fix Model union types that are objects, arrays, escaped characters

jasoncomes:
The components that utilize models can employ an arbitrary data type, which can be arrays, objects, or even escaped characters. To properly type this, we need to validate what these data types are and then return the correct corresponding types.

```
# duro-plm

SyntaxError: ';' expected. (15:17)
  13 |  * @example [{\&#34;con\&#34;: \&#34;asc\&#34;}]
  14 |  */
> 15 |   orderBy?: "[{"con": "asc"}]" | "[{"con": "desc"}]" | "[{"lastModified": "asc"}]" | "[{"lastModified": "desc"}]" | "[{"name": "asc"}]" | "[{"name": "desc"}]";
     |                 ^
  16 |    /**
  17 |  * First N Items
  18 |  * The number of items to return.
```

```
# text-manipulation

SyntaxError: Unterminated string literal. (18:18)
  16 |  * @placeholder The type of newline to use
  17 |  */
> 18 |   newlineType?: "
     |                  ^
  19 | " | "
  20 | ";
  21 |    /**
```

### DIFF
--- a/packages/spectral/src/generators/componentManifest/getInputs.ts
+++ b/packages/spectral/src/generators/componentManifest/getInputs.ts
@@ -89,7 +89,11 @@ export const INPUT_TYPE_MAP: Record<InputFieldDefinition["type"], ValueType> = {
 
 const getInputValueType = (input: ServerTypeInput) => {
   const valueType = input.model
-    ? input.model.map((choice) => `"${choice.value}"`).join(" | ")
+    ? input.model
+        .map((choice) => {
+          return `\`${choice.value.replaceAll("\r", "\\r").replaceAll("\n", "\\n")}\``;
+        })
+        .join(" | ")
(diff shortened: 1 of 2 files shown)


packages/spectral/src/generators/componentManifest/getInputs.ts
```diff
@@ -87,9 +87,24 @@ export const INPUT_TYPE_MAP: Record<InputFieldDefinition["type"], ValueType> = {
   dynamicFieldSelection: "string",
 };
 
+const ESCAPED_REPLACEMENTS: Record<string, string> = {
```
jasoncomes, line 90: I attempted a ton of different regex with replacement groups here, however a direct match was the only thing that worked.

ryanwersal (approved):

PR #254, opened 2024-07-26
updates convertOnExecution to pass along default values

awburgard:


packages/spectral/src/generators/utils/updateWithDefaultValues.ts
```diff
@@ -0,0 +1,24 @@
+// Provide default values at runtime if user did not provide any during CNI build
+export const updateWithDefaults = (
+  manifest: Record<string, any>,
+  inputValue: Record<string, any>,
+  convertType: "execution" | "reference",
```
awburgard, line 5: what the conversion for component reference and the execution need are slightly different, rather than two separate functions just wanted to have this check here

bkegley (approved):
Looks good!

PR #258, opened 2024-08-02
Make `key` optional to support previously published manifests

bkegley:


awburgard (approved):

jasoncomes (commented):
Do we also need to make this key optional?

```
export interface ComponentManifest {
  key?: string; <---
  public: boolean;
  signature: string | null;
  actions: Record<string, ComponentManifestAction>;
  triggers: Record<string, ComponentManifestTrigger>;
  dataSources: Record<string, ComponentManifestDataSource>;
  connections: Record<string, ComponentManifestConnection>;
}
```

bkegley:
That `key` is necessary and has been since the beginning of this.

PR #265, opened 2024-08-27
Component manifests - Escape input strings

pattra:
This PR addresses an issue where including `/* doc blocks */` or `` `backticked` `` values in input fields would cause the component manifest generator to fail.

### DIFF
--- a/packages/spectral/package.json
+++ b/packages/spectral/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "9.1.2",
+  "version": "9.1.3",
   "description": "Utility library for building Prismatic components",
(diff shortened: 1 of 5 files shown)


jasoncomes (commented):
Looking good, @pattra. However, these escaped values are causing build errors in the generated component manifests when you attempt to run `npm install` or `npm run build`. Tested on the Code Component.

Screen Shot 2024-08-28 at 12 47 11 PM

pattra:
@jasoncomes Thanks for catching this. Just updated the regex to look for `$` as well; then ran `component-manifest` on `code` and `mysql` and successfully built both manifests. Ready for re-review!

bkegley (approved):

PR #266, opened 2024-08-28
CNI - Bugfix - Default keyvaluelist items not converting to valid YAML

pattra:
This PR addresses a CNI issue where including default `keyvaluelist` items as the type hinting described would not build valid YAML. It includes:

* An update to `convertConfigVar` to enable both `keyvaluelist` and `valuelist` config vars to be processed properly
* Basic tests for config var input conversion


bkegley (approved):
This looks great!

PR #274, opened 2024-10-15
CNI - Support write-only inputs for built-in component connections

pattra:


packages/spectral/src/types/ComponentRegistry.ts
```diff
@@ -45,7 +45,8 @@ type ComponentReferenceTypeValueMap<
     actions: ValueExpression<TValue>;
     triggers: ValueExpression<TValue> | ConfigVarExpression;
     dataSources: ValueExpression<TValue> | ConfigVarExpression;
-    connections: (ValueExpression<TValue> | ConfigVarExpression) & ConfigVarVisibility;
+    connections: (ValueExpression<TValue> | ConfigVarExpression) &
+      ConfigVarVisibility & { writeOnly?: true };
```
jpeters5392, line 49: This seems wrong to me.  Connections themselves aren't marked as writeOnly, it is the individual inputs on a connection that get marked as writeOnly.

pattra: This type refers to the `connections.values` block, not the connection itself.

Screenshot 2024-10-17 at 2 31 46 PM

Hinting examples:
Screenshot 2024-10-17 at 2 35 59 PM
Screenshot 2024-10-17 at 2 37 00 PM

jpeters5392 (approved):

PR #275, opened 2024-10-22
Support componentized polling triggers

pattra:
# Summary
This PR allows component authors to create `pollingTriggers` that are optionally attached to an existing component action.

While in the `context` of a polling trigger's `perform` function, the dev will have access four new primitives:

1. `context.polling.invokeAction(actionParams)`: Invokes the `pollAction`, with type hinting for the action params
2. `context.polling.getState()`: Retrieves the contents of `instanceState.__prismaticInternal.polling`
3. `context.polling.setState(newState)`: Sets the contents of `instanceState.__prismaticInternal.polling`. Will stomp current contents.
4. `context.polling.reinvokeFlow(data, config)`: A wrapper around `axios.post` that calls `context.invokeUrl`. Can be used for complex pagination cases.

Note on 4: We may eventually introduce a general `invokeFlow` that can be used in other contexts and may access other flows. That functionality is not in scope for this feature.

# Code example

```ts
import {
  getPaginationFromResults,
  shouldReinvoke
} from "./somewhere";

type MyPaginationType = {
  limit?: number;
  page?: number;
};

const myPollingTrigger = pollingTrigger({
  pollAction: listProducts,
  perform: async (context, payload, params) => {
    const pollState: MyPaginationType = context.polling.getState();
    
    // Params will hint on both trigger inputs & associated action inputs
    const limit = pollState.limit ?? params.limit;
    const pageToFetch = pollState.page ?? params.page;

    // Will hint only on action inputs
    const results = context.polling.invokeAction({
      ...params,
      page: pageToFetch,
    });
    
    const newPollState = getPaginationFromResults(results);
    context.polling.setState(newPollState);

    if (shouldReinvoke(results)) {
      // if they don't want to use internalState, they could modify
      // the body or params here and then handle that in the perform.
      context.polling.reinvokeFlow(payload.body, payload.params);
    }
    
    return Promise.resolve({ payload: results });
  },
});
```

### DIFF
--- a/packages/spectral/src/types/PollingTriggerDefinition.ts
+++ b/packages/spectral/src/types/PollingTriggerDefinition.ts
@@ -1,3 +1,4 @@
+import type { AxiosRequestConfig, AxiosResponse } from "axios";
 import type {
   ActionDisplayDefinition,
   TriggerEventFunction,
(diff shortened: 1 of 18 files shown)


jpeters5392 (changes requested):
Marking as request changes until the types issues are resolved

packages/spectral/src/types/PollingTriggerDefinition.ts
```diff
@@ -0,0 +1,73 @@
...
+  Inputs,
+  TriggerResult,
+  ConfigVarResultCollection,
+  TriggerPayload,
+  ActionDefinition,
+  ActionContext,
+  ActionInputParameters,
+} from ".";
+
+export type PollingContext<TInputs extends Inputs = Inputs> = {
```
bkegley, line 14: Extending `ActionContext` here instead of using an intersection as the `context` type on line 33 will prevent inadvertent property overrides with future changes.

bkegley (approved):
This looks really good. Just a small change.

jpeters5392 (approved):

PR #278, opened 2024-11-05
Add default retryCondition method to AxiosRetryConfig

pattra:
Some users were ending up in a situation where leaving `retryCondition` undefined would result in a `retryCondition is not a function` error.

Normally leaving this field blank shouldn't cause issues (`retryCondition` is an optional field), but this fallback should cover any strange corner cases.


bkegley (approved):
LGTM!

PR #280, opened 2024-11-12
Run input clean functions even if value not explicitly provided

ryanwersal:
Previous behavior was to iterate over the collection of provided inputs but this meant that certain scenarios, such as version upgrades, wouldn't get an opportunity to run input clean functions if the value wasn't present in the collection of input values.

Now clean functions will be given `undefined` and an opportunity to run and clean the value into something usable/expected.

### DIFF
--- a/packages/spectral/package.json
+++ b/packages/spectral/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "10.1.1",
+  "version": "10.2.0",
   "description": "Utility library for building Prismatic components and code-native integrations",
   "keywords": ["prismatic"],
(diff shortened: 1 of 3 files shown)


packages/spectral/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "10.1.1",
+  "version": "10.2.0",
```
jpeters5392, line 3: No corresponding version bump in `package-lock.json`?

ryanwersal: I re-ran `yarn` and it didn't modify any lockfiles.

PR #298, opened 2025-01-10
Jp/fix lodash imports

jpeters5392:


pattra:
Looks good to me though your local formatting settings seem to differ from what's expected -- running `npm run format` in `spectral/packages/spectral` should fix?

pattra (approved):

PR #317, opened 2025-03-21
Support htmlElement configVars

pattra:
Previously CNI developers could not define `htmlElement` config vars in their config pages. This change makes it so the typing supports it & that they get converted into proper YAML (i.e. htmlElements should not show in the `requiredConfigVars` block)

### DIFF
--- a/packages/spectral/src/serverTypes/convertIntegration.ts
+++ b/packages/spectral/src/serverTypes/convertIntegration.ts
@@ -116,17 +116,24 @@ const convertConfigPages = (
     name,
     tagline,
     ...(userLevelConfigured ? { userLevelConfigured } : {}),
-    elements: Object.entries(elements).map(([key, value]) =>
-      typeof value === "string"
-        ? {
-            type: "htmlElement",
-            value,
-          }
-        : {
-            type: "configVar",
-            value: key,
-          },
-    ),
+    elements: Object.entries(elements).map(([key, value]) => {
+      if (typeof value === "string") {
+        return {
+          type: "htmlElement",
+          value,
+        };
+      } else if ("dataType" in value && value.dataType === "htmlElement") {
+        return {
+          type: "htmlElement",
+          value: key,
+        };
+      }
+
+      return {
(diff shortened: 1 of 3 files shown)


packages/spectral/src/serverTypes/convertIntegration.ts
```diff
@@ -116,17 +116,24 @@ const convertConfigPages = (
...
-            value: key,
-          },
-    ),
+    elements: Object.entries(elements).map(([key, value]) => {
+      if (typeof value === "string") {
+        return {
+          type: "htmlElement",
+          value,
+        };
+      } else if ("dataType" in value && value.dataType === "htmlElement") {
```
hipplej, line 125: Do we need to check if `value` is an object here?

packages/spectral/src/serverTypes/convertIntegration.ts
```diff
@@ -188,9 +195,13 @@ const codeNativeIntegrationYaml = (
     { ...(configVars ?? {}) },
   );
 
-  const requiredConfigVars = Object.entries(configVarMap).map(([key, configVar]) =>
-    convertConfigVar(key, configVar, referenceKey, componentRegistry),
-  );
+  const requiredConfigVars: Array<ServerRequiredConfigVariable> = [];
+
+  Object.entries(configVarMap).forEach(([key, configVar]) => {
+    if ("dataType" in configVar && configVar.dataType !== "htmlElement") {
```
hipplej, line 201: I guess same question here. I'm not sure if maybe we can safely implicitly assume that is is or not, not being super familiar with this code myself.

hipplej (approved):
Just a couple questions, but if we can safely assume the things are already `objects` then we are probably good to go so I'll just mark it approved.

packages/spectral/src/serverTypes/convertIntegration.ts
pattra: It should be but can add a check to be safe.

PR #321, opened 2025-03-27
Support importMetadata

pattra:
This PR supports adding build/import-specific metadata to generated CNI.

This should not be merged until the definition schemas are updated in the backend.


packages/spectral/src/serverTypes/convertIntegration.ts
```diff
@@ -55,6 +55,9 @@ import {
 import merge from "lodash/merge";
 import { createInvokeFlow } from "./perform";
 import { createDebugContext, logDebugResults } from "./context";
+import path from "path";
+import { readFileSync } from "fs";
```
hipplej, line 59: I get scolded every time I use one of these synchronous versions even though it likely makes _zero_ difference in reality. So I'm fine with this but be warned 😄

hipplej (approved):

PR #322, opened 2025-03-31
Add .vscode directory

taylorreece:
Without a .vscode directory, my VS Code's prettier and biome plugins fight with one another over formatting. This explicitly declares that biome is the formatter for this project (which it already is).


ryanwersal (approved):
🎉

PR #324, opened 2025-03-31
Support templateConnections

pattra:
This should not be merged until the corresponded backend work is in place, but I'm putting this up for now for early review.

## Example usage

```typescript
const myTemplateConnection = templateConnection({
  key: "my-template-connection",
  display: {
    label: "My Template Connection",
    description: "A test template connection",
  },
  inputs: {
    domain: {
      label: "Domain",
      type: "string",
      required: true,
    },
  },
  templateInputs: {
    authorizeUrl: {
      label: "Authorization URL",
      defaultValue: "https://{{#domain}}.myapp.com/authorizations/new",
    },
  },
});
```


packages/spectral/src/types/ConnectionDefinition.ts
```diff
@@ -54,6 +60,65 @@ export interface OnPremConnectionDefinition extends BaseConnectionDefinition {
...
+    : TConnectionType extends "client_credentials"
+      ? TTemplateComplement<OAuth2ClientCredentialInputs, TInputs> & {
+          [key: string]: ConnectionTemplateInputField;
+        }
+      : TTemplateComplement<OAuth2AuthorizationCodeInputs, TInputs> & {
+          [key: string]: ConnectionTemplateInputField;
+        };
+}) {
+  return merge(inputSet.inputs, inputSet.templateInputs);
+}
```
bkegley, line 121: In order to future-proof this against breaking changes we'll want this as multiple arguments instead of an object with `inputs` and `templateInputs` properties. When we do provide stronger typing for the available template tags based on the provided inputs it'll require that we know the `inputs` in order to infer the types for the `templateInputs`. In general, objects don't play as nicely with these kinds of progressive types since there's nothing stopping you from first providing the `templateInputs` property and then moving to `inputs`. In these cases it's better to use multiple params.

bkegley (commented):
This looks great. There's one change we'll want to make now so we don't need to introduce a breaking change when we improve the typing experience.

bkegley (approved):

PR #327, opened 2025-04-10
Add basic connection template input validation

pattra:
This PR adds basic validation for `connectionTemplateInput` template strings. We check that:

* Keys referenced in the template string must be declared as inputs
* Keys referenced in the template do not reference other template inputs
* The template string contains some number of template slots


hipplej (approved):
LGTM

PR #330, opened 2025-04-29
Support connection template inputs in CNI inline connection definitions

pattra:


ryanwersal (approved):

taylorreece (approved):
I tested this change and verified that connection inputs template properly in a code-native integration. Thanks for tackling that @pattra !

PR #335, opened 2025-05-16
Fix scoped config variable visibility

taylorreece:
Scoped config variables are treated specially in the configuration wizard. In the low-code designer, YAML definitions (left) do _not_ contain SCVs within `configPages.*.elements`. In code-native (right), they currently do. That's a mistake.

image

Similarly, in low-code (left) SCVs have `meta.*` visibility settings. In code-native (right) they don't.

image

This PR fixes both of those issues, so code-native integration adhere to the standard that low-code integrations use.

This resolves the issue where a user may see a red toaster `Inputs cannot be empty` alert when using scoped config variables in code native.

image

### DIFF
--- a/packages/spectral/src/serverTypes/convertIntegration.ts
+++ b/packages/spectral/src/serverTypes/convertIntegration.ts
@@ -129,29 +129,31 @@ const convertConfigPages = (
     name,
     tagline,
     ...(userLevelConfigured ? { userLevelConfigured } : {}),
-    elements: Object.entries(elements).map(([key, value]) => {
-      if (typeof value === "string") {
+    elements: Object.entries(elements)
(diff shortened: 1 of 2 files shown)


packages/spectral/src/serverTypes/convertIntegration.ts
```diff
@@ -624,6 +626,10 @@ export const convertConfigVar = (
       stableKey,
       dataType: "connection",
       useScopedConfigVar: stableKey,
+      meta: {
```
pattra, line 629: Is it safe to always set both of these values to false? Is there a need to check the original values here and carry them over?

pattra (approved):

PR #337, opened 2025-06-12
Standardize CNI `context.components` availability

pattra:
A bug was reported where, when defining `onInstanceDeploy` and `onInstanceDelete` methods for a CNI flow, `context.components` was unavailable.

This PR makes it so `context.components` is exposed in all CNI-related contexts. To achieve this I did a bit of refactoring, including:

* moving the logic for generating component invocation methods from `convertOnExecution` into a shared `createCNIContext` method, and
* moving `createInvokeFlow` from the component `perform` helper file into the `context` helper file.

### DIFF
--- a/packages/spectral/package.json
+++ b/packages/spectral/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "10.5.9",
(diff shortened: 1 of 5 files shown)


ryanwersal (approved):
LGTM 👍

PR #345, opened 2025-07-25
Bugfix: Update test harness connection keys

pattra:
A bug was reported where the `createConfigVars` testing method would not assign the correct keys to created connections. The issue was that `connection.key` was getting set to the connection's display key, rather than its actual key.

### DIFF
--- a/packages/spectral/src/testing.ts
+++ b/packages/spectral/src/testing.ts
@@ -380,7 +380,7 @@ export const invokeDataSource = async <
   return result;
 };
 
-type TestConnectionValue = Pick<ConnectionValue, "fields" | "context" | "token">;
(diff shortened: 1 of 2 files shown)


packages/spectral/src/testing.ts
```diff
@@ -399,7 +399,7 @@ const createConfigVars = <TConfigVarValues extends TestConfigVarValues>(
         ...result,
         [key]: {
           ...value,
-          key,
+          key: value.key,
```
ryanwersal, line 402: Wouldn't we accomplish the same mission by deleting this line since we're spreading `value`?

ryanwersal (approved):

PR #351, opened 2025-08-08
Enable action invocation via manifest

pattra:
This PR updates the component-manifest generator so that the `perform` methods are functionally equivalent to `context.components` invokes, and not just no-ops.

To do this, we make it so that management triggers and flow onExecutions invokes are wrapped with an `asyncLocalStorage` context that includes the action execution context. As a result, invokes like this (which an AI agent is likely to lean on once manifests are in-src) become valid:

```typescript
import { actions as slack } from "../manifests/slack";

// in some flow's onExecution...
  const postMessageResult2 = await slack.postMessage.perform({
    connection: context.configVars["Org Slack Connection"],
    message: "hi from special invoke",
    channelName: context.configVars["Slack Channel"],
  });
```

As far as I can tell there's no way to conditionally use `asyncLocalStorage` based on whether the invoke is of this type vs the `components.COMPONENT_KEY.ACTION_KEY` invocation, but I'm open to feedback.


packages/spectral/src/generators/componentManifest/createActions.ts
```diff
@@ -120,6 +121,7 @@ interface RenderActionProps {
     description: string;
     inputs: Input[];
     examplePayload?: unknown;
+    component: string;
```
bkegley, line 124: Could we name this `componentKey` for specificity? `component` often refers to the combination of key, version, and public.

packages/spectral/src/generators/componentManifest/templates/actions/action.ts.ejs
```diff
@@ -12,9 +13,12 @@ export interface <%= action.typeInterface %>Values {
...
 export const <%= action.import %> = {
   key: "<%= action.key %>",
-  perform: <TReturn>(
-    _values: <%= action.typeInterface %>Values
-  ): Promise<TReturn> => Promise.resolve<TReturn>({} as TReturn),
+  perform: async <TReturn>(
+    values: <%= action.typeInterface %>Values
+  ): Promise<TReturn> => {
+    const context = requireContext();
+    return await context.components.<%= action.component %>.<%= action.key %>({ ...values }) as TReturn;
```
bkegley, line 20: Just to make sure I'm thinking about this correctly - we don't need to worry about existing _published_ component manifests which don't have `action.component` defined following this code path because this will only be generated for net-new components and only exposed as a callable function (at least reasonably discoverable) for inline generated manifest. Is that correct?

packages/spectral/src/serverTypes/asyncContext.ts
```diff
@@ -0,0 +1,23 @@
+import { AsyncLocalStorage } from "node:async_hooks";
+import { ActionContext } from "../types";
+
+const actionContextStorage = new AsyncLocalStorage<ActionContext>();
+
+export function getCurrentContext(): ActionContext | undefined {
+  return actionContextStorage.getStore();
+}
```
bkegley, line 8: Is there a reason we'd want to allow calling this without the `requireContext` checks or should we inline this in the `requireContext` function?

packages/spectral/src/generators/componentManifest/createActions.ts
pattra: Key is enough here. Will name it more specifically

bkegley (approved):

PR #353, opened 2025-08-12
Remove clean function from ConnectionInput hinting

pattra:
We do not currently support defining `clean` functions on connection inputs. While `clean` is fine for every other type of input, it's basically because all other cases involve a `perform` function that we can modify to run the clean functions at the convert step.

Connections do not have a `perform` step, and at component build time they're basically defined as objects. Because of this, we cannot run `clean` methods on their inputs without doing some significant investigation re: storing the `clean` functions somewhere and deciding where they'll be executed (the runner?).

That said, we still show an optional `clean?` key in connection input hinting, which has caused confusion. This PR removes the incorrect hinting.

### DIFF
--- a/packages/spectral/src/serverTypes/convertComponent.ts
+++ b/packages/spectral/src/serverTypes/convertComponent.ts
@@ -38,7 +38,7 @@ export const convertInput = (
     label,
     collection,
     ...rest
-  }: InputFieldDefinition | OnPremConnectionInput,
+  }: InputFieldDefinition | OnPremConnectionInput | ConnectionInput,
 ): ServerInput => {
   const keyLabel =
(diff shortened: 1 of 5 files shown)


packages/spectral/src/serverTypes/convertComponent.ts
```diff
@@ -286,7 +286,7 @@ export const convertConnection = ({
 
   const convertedInputs = Object.entries(inputs).map(([key, value]) => {
     if ("templateValue" in value) {
-      return convertTemplateInput(key, value, inputs);
+      return convertTemplateInput(key, value as ConnectionTemplateInputField, inputs);
```
ryanwersal, line 289: What necessitates the cast? Are we able to update `convertTemplateInput` to make the types make sense?

ryanwersal (approved):

PR #356, opened 2025-08-22
Organize spectral imports/exports

pattra:
This PR makes it so that commonly imported files from spectral (e.g. types, component dev fns, integration dev fns) are a bit more optimized and will hopefully not drag in a ton of other dependencies with them.

Dev specific imports can be accessed via:
```
@prismatic-io/spectral/dist/component
@prismatic-io/spectral/dist/integration
```

And exist imports to `@prismatic-io/spectral/dist/types` and the like should be improved.


packages/spectral/biome.jsonc
```diff
@@ -48,5 +48,26 @@
...
+  "overrides": [
+    {
+      "include": ["src/types/**/*"],
+      "linter": {
+        "rules": {
+          "nursery": {
+            "noRestrictedImports": {
+              "level": "error",
+              "options": {
+                "paths": {
```
pattra, line 61: A bit silly but AFAIK there is no `pattern` option here, just `paths`?

packages/spectral/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "10.6.1",
+  "version": "10.7.0",
```
pattra, line 3: Let me know if we think this should be major.

packages/spectral/biome.jsonc
jpeters5392: It was added in 2.2.0 of biome https://biomejs.dev/linter/rules/no-restricted-imports/#patterns

packages/spectral/package.json
jpeters5392: I think this should probably be a major given it will break certain consumers.

packages/spectral/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@prismatic-io/spectral",
-  "version": "10.6.3",
+  "version": "10.6.4",
```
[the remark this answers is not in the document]
pattra: We swapped it back to being patch because there are no longer breaking changes being introduced after the latest reorganization

PR #358, opened 2025-08-27
Transform queueConfig.dedupeIdField into proper shape for YAML generation

pattra:


taylorreece (approved):
Latest works like a charm; thanks!

PR #369, opened 2025-09-22
Bugfix: cni-component-manifest - Throw readable error if there is an API request failure

pattra:
Tested by setting the access token to an intentionally invalid string. Most of the diff here is just formatting changes from putting stuff into a try/catch block.


bkegley (approved):
That's a very unfortunate diff. But LGTM

PR #432, opened 2026-04-14
Update dev dependencies

ryanwersal:
Most churn here is biome induced. This migrates to Biome 2 and homogenizes the configs. I separated out the autofix/autoformat into a separate to attempt to ease review.


packages/spectral/package.json
```diff
@@ -49,7 +51,7 @@
     "jest-mock": "29.7.0",
     "lodash": "^4.18.1",
     "object-sizeof": "^2.6.4",
-    "prettier": "2.6.2",
+    "prettier": "^3.8.2",
```
bkegley, line 54: Is `prettier` solely used for formatting generated code now?

bkegley (approved):

packages/spectral/package.json
ryanwersal: Should be. Are you interested in swapping it to Biome? Could land that as a separate body of work. I'm definitely onboard with getting off prettier across the board. 💯

bkegley: For the `embedded` typegen I used biome but that was net-new functionality. Wouldn't hate completely ditching prettier.

PR #443, opened 2026-04-15
Update biome/prettier; tweak dependabot config

ryanwersal:
Try to roll version updates into an MR-per-week but let CVE fixes pop individually.


katrynmc (approved):
Let's try it! I think we may still want to exclude majors because of the difference in risk and review complexity but depending on the codebase and library utilization that may not be so problematic.

PR #455, opened 2026-04-28
Add structuredObject input type and structuredObjectInput wrapper

CaptainChemist:
# Summary

Adds a new `structuredObject` input field type and a `structuredObjectInput()` factory function to the spectral SDK. Component authors can now group related primative inputs under a named container that the platform's component publishing accepts.

Example:
```
import { input, structuredObjectInput } from "@prismatic-io/spectral";

inputs: {
  name: structuredObjectInput({
    label: "Name",
    inputs: {
      first: input({ type: "string", label: "First Name", required: true }),
      last: input({ type: "string", label: "Last Name", required: true }),
      prefix: input({ type: "string", label: "Prefix" }),
    },
  }),
  email: input({ type: "string", label: "Email", required: true }),
}
```

# Changes

- New `StructuredObjectInputField` type added to `InputFieldDefinition` union. Has a `type` of `structuredObject` and a recursive `inputs: Record<string, LeafInputFieldDefinition>` member.
- New `LeafInputFieldDefinition` type alias is the union of all non-structuredObject members. Used as the children's allowed type to flag attempts to nest a structuredObject inside a structuredObject. This is due to a depth of 1 cap the api enforces.
- The `Input` interface gains an optional `inputs?: Input[]` field. Mirrors recursive `InputFieldDefinition` graphQL input type that the platform's `publishComponent` mutation accepts.
- `InputFieldDefaultMap` is extended with `structuredObject: unknown` to be circled back with in a later CNI update to add support for these nested objects.
- `ActionInputParameters` also maps `structuredObject` to `unknown` which will also be resolved in a later CNI update.

### DIFF
--- a/packages/spectral/src/component-testing.test.ts
+++ b/packages/spectral/src/component-testing.test.ts
@@ -1,5 +1,5 @@
 import { describe, expect, it } from "vitest";
-import { action, input, pollingTrigger, trigger } from ".";
+import { action, component, input, pollingTrigger, structuredObjectInput, trigger } from ".";
 import { convertTrigger } from "./serverTypes/convertComponent";
 import { PerformFn } from "./serverTypes/perform";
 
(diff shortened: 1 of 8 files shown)


packages/spectral/src/serverTypes/convertComponent.ts
```diff
@@ -29,21 +29,42 @@ import {
   Input as ServerInput,
   Trigger as ServerTrigger,
 } from ".";
-import { createPerform, createPollingPerform, InputCleaners, PerformFn } from "./perform";
+import { CleanFn, createPerform, createPollingPerform, InputCleaners, PerformFn } from "./perform";
+
+const cleanerFor = (input: InputFieldDefinition): CleanFn | undefined =>
```
robertmaloney, line 34: Do we need to cast `input.clean` as potentially undefined?

packages/spectral/src/serverTypes/convertComponent.ts
```diff
@@ -185,8 +207,16 @@ export const convertTrigger = <
...
+    // Cast: the type-guard call's narrowing doesn't survive being stored in
+    // `isPollingTrigger` and read here.
+    const { pollAction: action } = trigger as PollingTriggerDefinition<
+      Inputs,
+      ConfigVarResultCollection,
+      TriggerPayload,
+      boolean,
+      any,
+      any
+    >;
```
robertmaloney: Can we use one of these generics to prevent the `InputFieldDefinition` case below?

packages/spectral/src/types/ActionInputParameters.ts
```diff
@@ -5,21 +5,33 @@ import type {
...
 } from "./Inputs";
 
+/** Resolves a single InputFieldDefinition's runtime value type. structuredObject
+ * resolves to `unknown` until the per-field record-type recursion lands. */
+type InputValue<T> = T extends StructuredObjectInputField
+  ? unknown
+  : T extends { clean: InputCleanFunction<any> }
+    ? ReturnType<T["clean"]>
+    : T extends { type: "connection"; collection?: infer C }
+      ? ExtractValue<Connection, C extends InputFieldCollection | undefined ? C : undefined>
```
robertmaloney: This should only check extension of `InputFieldCollection` no? It seems this `infer C` pattern is getting funky or misbehaving compared to the old way.

packages/spectral/src/types/Inputs.ts
```diff
@@ -380,6 +382,36 @@ export type DateTimeInputField = BaseInputField & {
   clean?: InputCleanFunction<unknown>;
 } & CollectionOptions<string>;
 
+/** `InputFieldDefinition` minus `StructuredObjectInputField`; used to cap
+ * structuredObject nesting at one level. */
+export type LeafInputFieldDefinition =
```
robertmaloney: Can we derive this from `InputFieldDefinition` and just omit `StructuredObjectInputField`?

packages/spectral/src/component-testing.test.ts
```diff
@@ -201,3 +201,67 @@ describe("test convert trigger", () => {
     }
   });
 });
+
+describe("structuredObject inputs in a published component", () => {
```
robertmaloney, line 205: Do we have type checking tests anywhere that setting type to something other than structuredObject won't transpile?

packages/spectral/src/serverTypes/convertComponent.ts
CaptainChemist: You are right, I can remove that since `CleanFn` can be `undefined` anyway and it all works.

packages/spectral/src/serverTypes/convertComponent.ts
CaptainChemist: Yes, but I have to move `isPollingTriggerDefinition` to be called within `if` so we get proper type narrowing. I have to do this again in the second place we used `isPollingTrigger`. Still the right call I think.

packages/spectral/src/types/ActionInputParameters.ts
CaptainChemist: Good catch! If we constrain `collection?: InputFieldCollection` directly in the `extends` shape, `T["collection"]` then becomes assignable to `ExtractValue`'s parameter without the `infer C` pattern.

packages/spectral/src/types/Inputs.ts
CaptainChemist: Great idea- switched this over. Much cleaner!

packages/spectral/src/component-testing.test.ts
CaptainChemist: I just added at `@ts-expect-error` test for the nested structuredObject case. I noticed when I made this case that the `structuredObjectInput` type: `T extends Omit<StructuredObjectInputField, "type">` as written allowed passing in a stray type key without giving an error on compile. For example:
```
const name = structuredObjectInput({
  label: "Name",
  type: "string", 
  inputs: {
    first: input({ type: "string", label: "First Name" }),
  },
});
```

I added `& { type?: never; }` to the end of the type so this will correctly error out if someone tries to set the type.

robertmaloney (approved):

PR #457, opened 2026-04-30
Replace uuid package with crypto.randomUUID

ryanwersal:
We only had the one callsite and the library was causing issues with our current jest-mock test harness.


jakethagle (approved):
LGTM!

PR #459, opened 2026-05-07
Config vars should have configurable scheduleType

katrynmc:
CNI never exposed scheduleType to authors and hard-coded "custom" in the converter, even though the platform has always accepted the full `"none" | "custom" | "minute" | "hour" | "day" | "week"` set — so this fix just plumbs scheduleType through to the converter, with a fallback inferred from defaultValue.


pattra (approved):
LGTM, though I'm not sure if Ryan's hit this repo yet with an updated release process? Do we still need to bump version here?

katrynmc:
@pattra it seems like it has landed so the version bump will happen on publish now!

PR #461, opened 2026-05-12
structuredObject perform() type fix

CaptainChemist:
What
This is `perform()` side typing improvement for `structuredObject`. It replaces the unknown placeholder in `InputValue<T>` for `StructuredObjectInputField` with a recursive mapped type that resolves to the declared children's value types.

Currently if we create a contact like this:

```
const createContact = action({
  inputs: {
    name: structuredObjectInput({
      label: "Name",
      inputs: {
        first: input({ type: "string", label: "First", required: true }),
        last:  input({ type: "string", label: "Last",  required: true }),
      },
    }),
    email: input({ type: "string", label: "Email" }),
  },
  perform: async (context, params) => { ... },
});
```
we will see `params.name` resolve to `unknown`.

After: `params.name` is `{ first: unknown; last: unknown }`. The parent shape is exposed, undeclared keys are rejected:

```
perform: async (context, params) => {
  params.name.first;         // ✓ accessible (resolves to unknown — see below)
  params.name.last;          // ✓ accessible
  params.name.middle;        // ✗ TS error: Property 'middle' does not exist
};
```

### DIFF
--- a/packages/spectral/src/component-testing.test.ts
+++ b/packages/spectral/src/component-testing.test.ts
@@ -284,4 +284,27 @@ describe("structuredObject inputs in a published component", () => {
       },
     });
   });
+
+  it("exposes structuredObject params as a record keyed by declared children inside perform", () => {
+    action({
+      display: { label: "Create Contact", description: "Create a new contact" },
+      inputs: {
+        name: structuredObjectInput({
+          label: "Name",
+          inputs: {
+            first: input({ type: "string", label: "First Name", required: true }),
+            last: input({ type: "string", label: "Last Name", required: true }),
+          },
+        }),
+      },
+      perform: async (_context, params) => {
+        // Declared children are accessible (would error if params.name were `unknown`).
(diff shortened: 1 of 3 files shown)


packages/spectral/src/component-testing.test.ts
```diff
@@ -284,4 +284,27 @@ describe("structuredObject inputs in a published component", () => {
       },
     });
   });
+
+  it("exposes structuredObject params as a record keyed by declared children inside perform", () => {
```
ryanwersal: Types tests are probably best suited for the `spectral-test` package - we have a `tsd` based pipeline established there.

CaptainChemist: Thanks Ryan, I missed that we have tests over in that package. Moved!

robertmaloney (approved):

robertmaloney:
I think the new test might be erroring in an unexpected way, but this should be good once that check passes!

PR #462, opened 2026-05-13
Add CNI support for configuring triggers for batching

internetlee:
Introduces resolver support for triggers to enable:
- on deploy triggers for initial executions (backfills)
- trigger resolution to enable batching
- and the ability for both to be present on one trigger so they can feed the same flow


misteroneill (approved):

internetlee:
Resolved conflicts and pulled back the scope of this change to exclude on-deploy resolver behavior.

pattra (approved):

pattra:
✅

PR #464, opened 2026-05-14
Align tsd usage with best practices

ryanwersal:
Notably to avoid using tsc as that's what motivated a lot of extra comments about disabling tsc and telling biome to allow _those_ comments.


CaptainChemist (approved):
Looks good to me!

PR #474, opened 2026-05-20
Component manifests should use static typescript version

ryanwersal:
Instead of updating to align with Spectral's since we should have compat with the type defs themselves.


CaptainChemist (approved):
This worked great in my local testing, thank you!

PR #476, opened 2026-05-20
Recursively clean structuredObject and dynamicObject inputs

robertmaloney:
Auto-generate cleaners for these containers during component conversion so each child's clean function runs at runtime. structuredObject walks its declared inputs; dynamicObject walks the selected configuration's inputs (including nested structuredObject children). Developers do not declare clean on these container types — the conversion always supplies the cleaner.


CaptainChemist (approved):
Really illustrative tests, thanks I learned a lot here. Looks good to me!

PR #484, opened 2026-06-09
feat: add outputSchema to action definitions

CaptainChemist:
# What
Adds an optional `outputSchema` to action definitions, letting component authors declare the shape of an action's output `data` as a JSON Schema. 

`outputSchema` is a discriminated union:

```ts
type ActionOutputSchema    = { type: "actionOutput";    schema: JsonSchema }
type BranchingOutputSchema = { type: "branchingOutput"; branchSchemas: Record<string, JsonSchema> }
type OutputSchema = ActionOutputSchema | BranchingOutputSchema
```

This is descriptive-only, its emitted to the component manifest and consumed by the Designer's reference picker. It is not read or validated against at runtime.

### DIFF
--- a/packages/spectral/src/generators/componentManifest/createActions.ts
+++ b/packages/spectral/src/generators/componentManifest/createActions.ts
@@ -1,6 +1,6 @@
 import path from "path";
 import type { ConfigVarResultCollection, Inputs, TriggerPayload, TriggerResult } from "../../types";
-import type { JsonSchema } from "../../types/jsonforms/JsonSchema";
+import type { OutputSchema } from "../../types/OutputSchema";
 import type { ComponentForManifest } from "../cniComponentManifest/types";
 import { createImport } from "../utils/createImport";
 import { createTemplate } from "../utils/createTemplate";
@@ -150,7 +150,7 @@ interface RenderActionProps {
     description: string;
     inputs: Input[];
     examplePayload?: unknown;
-    outputSchema?: JsonSchema;
+    outputSchema?: OutputSchema;
     componentKey: string;
   };
   dryRun: boolean;
--- a/packages/spectral/src/generators/componentManifest/createActions.ts
+++ b/packages/spectral/src/generators/componentManifest/createActions.ts
@@ -1,7 +1,7 @@
 import path from "path";
 import type { ConfigVarResultCollection, Inputs, TriggerPayload, TriggerResult } from "../../types";
-import type { ComponentForManifest } from "../cniComponentManifest/types";
 import type { JsonSchema } from "../../types/jsonforms/JsonSchema";
+import type { ComponentForManifest } from "../cniComponentManifest/types";
 import { createImport } from "../utils/createImport";
 import { createTemplate } from "../utils/createTemplate";
 import { createTypeInterface } from "../utils/createTypeInterface";
@@ -91,7 +91,7 @@ export const createActions = async <
           description: action.display.description,
           inputs,
           ...(action.examplePayload ? { examplePayload: action.examplePayload } : {}),
-          ...(action.output ? { output: action.output } : {}),
+          ...(action.outputSchema ? { outputSchema: action.outputSchema } : {}),
           componentKey: component.key,
         },
         imports,
@@ -150,7 +150,7 @@ interface RenderActionProps {
     description: string;
     inputs: Input[];
     examplePayload?: unknown;
-    output?: JsonSchema;
+    outputSchema?: JsonSchema;
     componentKey: string;
   };
   dryRun: boolean;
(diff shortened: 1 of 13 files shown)


packages/spectral/src/types/OutputSchema.ts
```diff
@@ -0,0 +1,38 @@
...
+ * NOT supported with `dynamicBranchInput`, where branch names are computed at
+ * runtime and cannot be enumerated; publishing such a component is rejected
+ * server-side. (Unrelated to `dynamicObject` *inputs*, whose `configurations`
+ * are fully declared at build time — that's an input feature and does not
+ * affect `outputSchema`.)
+ */
+export interface BranchingOutputSchema {
+  type: "branchingOutput";
+  branchSchemas: Record<string, JsonSchema>;
+}
```
jasoncomes, line 31: Doesn't this conflict with what the backend expects? the API input type uses `schemas: [{ name, schema: JSONString }]`, array format not a record. Also the schema values need to be stringified (JSONString) before sending. Is prism handling that conversion somewhere or does this need to align to the API shape?

packages/spectral/src/types/ActionDefinition.ts
```diff
@@ -52,4 +53,18 @@ export interface ActionDefinition<
   dynamicBranchInput?: string;
   /** An example of the payload output by this action. */
   examplePayload?: Awaited<ReturnType<this["perform"]>>;
+  /**
+   * Declares the shape of this action's output `data` as a JSON Schema, used by
+   * the Prismatic UI to let integration authors reference this step's output
+   * before a real execution has produced data. A discriminated union:
+   * `{ type: "actionOutput", schema }` for a single payload shape, or
+   * `{ type: "branchingOutput", branches }` for a per-branch map of shapes.
+   * Descriptive only — it is not enforced at runtime.
```
jasoncomes, line 62: branches should be schemas here

packages/spectral/src/types/index.ts
```diff
@@ -28,6 +28,7 @@ export * from "./Inputs";
 export * from "./InstanceAttributes";
 export * from "./IntegrationAttributes";
 export * from "./IntegrationDefinition";
+export * from "./OutputSchema";
 export * from "./PollingTriggerDefinition";
 export * from "./ScopedConfigVars";
 export * from "./ScopedConfigVars";
```
jasoncomes: Not part of your MR, but noticed this is duplicated. Andd... its an easy cleanup lol

packages/spectral/src/generators/componentManifest/createActions.ts
```diff
@@ -90,6 +91,7 @@ export const createActions = async <
           description: action.display.description,
           inputs,
           ...(action.examplePayload ? { examplePayload: action.examplePayload } : {}),
+          ...(action.outputSchema ? { outputSchema: action.outputSchema } : {}),
           componentKey: component.key,
```
jasoncomes, line 94: Is this going to serialize correctly when prism publishes it? The manifest would have branchSchemas with parsed objects but the api expects a schemas array with stringified values

packages/spectral/src/types/ActionDefinition.ts
jasoncomes: Let me know what one we use! branchSchemas or branches or schemas? cc @robertmaloney

CaptainChemist: Sorry forgot to update this note to match `branchSchemas` like the rest:
`{ type: "branchingOutput", branchSchemas }`
fixed!

packages/spectral/src/types/index.ts
CaptainChemist: I only see one `./OutputSchema` export in this file, that's the only line that's changed in this MR. Maybe some github visualization issues?

packages/spectral/src/generators/componentManifest/createActions.ts
robertmaloney: Yeah it doesn't look like we have the convert layer implemented yet which should handle serializing the JSON schema objects.

packages/spectral/src/types/OutputSchema.ts
robertmaloney: This interface is fine as I believe it's used in spectral-land. But you're correct that the convert layer is needed.

packages/spectral/src/types/index.ts
robertmaloney: I think he's referring to the double `export * from "./ScopedConfigVars";`

CaptainChemist: Gah so sorry my bad! I'll get that fixed.

packages/spectral/src/types/OutputSchema.ts
CaptainChemist: Thanks for calling this out. Adding a converter in convertComponent to stringify the JSON before sending and it works great. I verified by publishing a component with a branching action, seeing it is stringified in the graphQL mutation sent by prism, and then reading it back over in the jethflow-null repo `sj/example-components` seems to work well.

packages/spectral/src/generators/componentManifest/createActions.ts
CaptainChemist: This branch has been updated and verified that the serialization process with prism works correctly.

jasoncomes (approved):

PR #493, opened 2026-06-23
Add spectral support for inline action calling

CaptainChemist:
# What

This adds support on the spectral side for inline action calling which lets the integration builder invoke an action's perform (or mock experimentalExamplePerform) from the Designer/EWB to populate live reference data.

# Changes

Author Facing type changes
- `ExperimentalPerformSupport` exposed along with a const companion which allows for authors to use either `ExperimentalPerformSupport.SAFE` or the string `"SAFE"` and get typechecking with both.
- Values are `"SAFE" | "UNSAFE" | "NOT_ALLOWED"` to match the backend graphQL enum type.

Publish-wire conversion
- `convertAction` wraps `experimentalExamplePerform` via`createPerform` to add input cleaning and error handling and mirroring `perform`. Its destructured out of the `...action` spread so the raw function doesn't leak onto the wire when it isn't wrapped. 
- The published `Action` wire type declares `experimentalExamplePerform` & both `experimentalExamplePerformSupport` and `experimentalPerformSupport` enums.

Component manifest
- `ComponentManifestAction` adds the two `*Support` scalars into the manifest so the frontend knows they are inline invokable. 
- `experimentalExamplePerform` is added to `ComponentManifestAction` type only. The generator doesn't emit a value or stub for it because its invoked by the published server action rather than the manifest.

CNI Component Manifest
- Threads the two `*Support` flags through the graphQL fetch path so components fetched from the API surface them.

### DIFF
--- a/packages/spectral/src/generators/cniComponentManifest/index.ts
+++ b/packages/spectral/src/generators/cniComponentManifest/index.ts
@@ -151,6 +151,12 @@ export const fetchComponentDataForManifest = async <
           },
           inputs: transformInputNodes(node.inputs.nodes),
           examplePayload: node.examplePayload,
+          ...(node.experimentalExamplePerformSupport
+            ? { experimentalExamplePerformSupport: node.experimentalExamplePerformSupport }
+            : {}),
+          ...(node.experimentalPerformSupport
+            ? { experimentalPerformSupport: node.experimentalPerformSupport }
+            : {}),
         };
       }
     });
@@ -265,6 +271,8 @@ async function getComponentActions(componentId: string, prismaticUrl: string, ac
                 }
               }
               examplePayload
+              experimentalExamplePerformSupport
+              experimentalPerformSupport
             }
             pageInfo {
               hasNextPage
--- a/packages/spectral/src/generators/cniComponentManifest/index.ts
+++ b/packages/spectral/src/generators/cniComponentManifest/index.ts
@@ -151,12 +151,8 @@ export const fetchComponentDataForManifest = async <
           },
           inputs: transformInputNodes(node.inputs.nodes),
           examplePayload: node.examplePayload,
-          ...(node.examplePerformSupport
-            ? { examplePerformSupport: node.examplePerformSupport }
-            : {}),
-          ...(node.performSupport
-            ? { performSupport: node.performSupport }
-            : {}),
+          ...(node.examplePerformSafety ? { examplePerformSafety: node.examplePerformSafety } : {}),
+          ...(node.performSafety ? { performSafety: node.performSafety } : {}),
         };
       }
     });
@@ -271,8 +267,8 @@ async function getComponentActions(componentId: string, prismaticUrl: string, ac
                 }
               }
               examplePayload
-              examplePerformSupport
-              performSupport
+              examplePerformSafety
+              performSafety
             }
             pageInfo {
               hasNextPage
--- a/packages/spectral/src/generators/cniComponentManifest/index.ts
+++ b/packages/spectral/src/generators/cniComponentManifest/index.ts
@@ -151,11 +151,11 @@ export const fetchComponentDataForManifest = async <
           },
           inputs: transformInputNodes(node.inputs.nodes),
           examplePayload: node.examplePayload,
-          ...(node.experimentalExamplePerformSupport
-            ? { experimentalExamplePerformSupport: node.experimentalExamplePerformSupport }
+          ...(node.examplePerformSupport
+            ? { examplePerformSupport: node.examplePerformSupport }
             : {}),
-          ...(node.experimentalPerformSupport
-            ? { experimentalPerformSupport: node.experimentalPerformSupport }
+          ...(node.performSupport
+            ? { performSupport: node.performSupport }
             : {}),
         };
       }
@@ -271,8 +271,8 @@ async function getComponentActions(componentId: string, prismaticUrl: string, ac
                 }
               }
               examplePayload
-              experimentalExamplePerformSupport
-              experimentalPerformSupport
+              examplePerformSupport
+              performSupport
             }
             pageInfo {
               hasNextPage
--- a/packages/spectral/src/serverTypes/convertComponent.test.ts
+++ b/packages/spectral/src/serverTypes/convertComponent.test.ts
@@ -10,6 +10,7 @@ import {
 } from "..";
 import {
   cleanerFor,
+  convertAction,
   convertConnection,
   convertInput,
   convertTemplateInput,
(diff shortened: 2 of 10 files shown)


packages/spectral/src/types/ActionDefinition.ts
```diff
@@ -4,6 +4,29 @@ import type { ComponentManifestAction } from "./ComponentManifest";
...
+ * author attests it is safe to run as-is; `UNSAFE` means running it has side effects /
+ * is not safe to invoke automatically; `NOT_ALLOWED` opts out entirely.
+ *
+ * Values are SCREAMING_SNAKE_CASE to match the backend GraphQL enum member names
+ * verbatim — these strings are published and read back through GraphQL, where a
+ * graphene Enum serializes by member name (not value), and spectral does no enum
+ * value mapping. Experimental — the name set and semantics may change before this graduates.
+ *
+ * Exposed as a const companion so authors can reference `ExperimentalPerformSupport.SAFE`
+ * instead of the raw string; the plain literal `"SAFE"` is still accepted everywhere.
```
robertmaloney: This seems overly verbose.

packages/spectral/src/types/ComponentManifest.ts
```diff
@@ -22,6 +23,16 @@ interface BaseInput {
 export interface ComponentManifestAction {
   key?: string;
   perform: (values: any) => Promise<unknown>;
+  /**
+   * Present only on manifests whose source carries an example perform. The generated
+   * manifest does not synthesize a delegating stub for it (there is no runtime
+   * `context.components` channel to delegate to, unlike `perform`); the example perform
+   * is invoked via the published server action. The `*Support` flags are the part the
+   * frontend reads from the manifest.
+   */
```
robertmaloney: Also seems overly verbose

packages/spectral/src/serverTypes/convertComponent.ts
```diff
@@ -388,6 +393,17 @@ const convertAction = (
       inputCleaners,
       errorHandler: hooks?.error,
     }),
+    // Wrap the example perform the same way as `perform` (input cleaning + error
+    // handling). Destructured out of `...action` so a raw, unwrapped function never
+    // leaks onto the server action; only emitted when the author defines it.
```
robertmaloney: Probably unnecessary comment.

packages/spectral/src/serverTypes/convertComponent.ts
```diff
@@ -388,6 +393,17 @@ const convertAction = (
...
+    ...(experimentalExamplePerform
+      ? {
+          experimentalExamplePerform: createPerform(experimentalExamplePerform, {
```
robertmaloney: Are we creating a problem by prefixing `experimental` here? We really want to avoid having to migrate component code or the DB in any way if we can.

CaptainChemist: I agree this could be a problem but I think its workable if we want to keep the experimental prefix. I'm seeing two switching costs here:

1) Name changing on the backend: If we name the DB column without the experimental prefix, we can make the graphQL field name different from the DB column name and surface the experimental prefix only at the api. So this seems straightforward to support.

2) Component author code: The problem here is the component author writes `experimentalPerformSupport: "SAFE"` and then this could break later if we remove the experimental prefix. We can keep the old experimental prefix to make it non breaking and set it as `@deprecated`. This creates more of an issue for the customers but its small and can be managed.

What do you think? Is the ceremony around experimental worth it here?

packages/spectral/src/serverTypes/convertComponent.ts
CaptainChemist: Thanks, removed!

packages/spectral/src/types/ComponentManifest.ts
CaptainChemist: Tightened.

packages/spectral/src/types/ActionDefinition.ts
CaptainChemist: You are totally right, tightened up.

packages/spectral/src/serverTypes/convertComponent.test.ts
```diff
@@ -28,6 +37,70 @@ describe("convertConnection", () => {
...
+    const converted = convertAction(
+      "doThing",
+      action({
+        display: baseDisplay,
+        inputs: {
+          count: input({ type: "string", label: "Count", clean: (v) => Number(v) }),
+        },
+        perform: async () => ({ data: null }),
+        experimentalExamplePerform: examplePerform,
+        experimentalExamplePerformSupport: "SAFE",
```
jasoncomes:
```suggestion
        experimentalExamplePerformSupport: ExperimentalPerformSupport.SAFE,
```

packages/spectral/src/generators/cniComponentManifest/index.ts
```diff
@@ -151,6 +151,12 @@ export const fetchComponentDataForManifest = async <
           },
           inputs: transformInputNodes(node.inputs.nodes),
           examplePayload: node.examplePayload,
+          ...(node.experimentalExamplePerformSupport
+            ? { experimentalExamplePerformSupport: node.experimentalExamplePerformSupport }
+            : {}),
+          ...(node.experimentalPerformSupport
+            ? { experimentalPerformSupport: node.experimentalPerformSupport }
+            : {}),
```
jasoncomes: So are we calling this ExampePerformSupport, have we thought on InlinePerformSupport|InlinePerform or similar so the ExamplePayload and ExamplePerform don't confuse the user?

packages/spectral/src/serverTypes/convertComponent.test.ts
```diff
@@ -28,6 +37,70 @@ describe("convertConnection", () => {
...
+      action({
+        display: baseDisplay,
+        inputs: {},
+        perform: async () => ({ data: null }),
+        experimentalPerformSupport: ExperimentalPerformSupport.UNSAFE,
+        experimentalExamplePerformSupport: ExperimentalPerformSupport.NOT_ALLOWED,
+      }),
+    );
+
+    expect(converted.experimentalPerformSupport).toBe("UNSAFE");
```
jasoncomes:
```suggestion
    expect(converted.experimentalPerformSupport).toBe(ExperimentalPerformSupport.UNSAFE);
```
Should these use the enums to test against, just curious. I'd ask claude

packages/spectral/src/serverTypes/convertComponent.test.ts
CaptainChemist: Great suggestion, I applied this manually after the type was changed `ExperimentalPerformSupport -> PerformSupport`.

packages/spectral/src/types/ActionDefinition.ts
```diff
@@ -4,6 +4,20 @@ import type { ComponentManifestAction } from "./ComponentManifest";
 import type { ActionDisplayDefinition } from "./DisplayDefinition";
 import type { ConfigVarResultCollection, Inputs } from "./Inputs";
 
+/**
+ * Whether an action may be invoked inline to populate reference data in the Designer/EWB:
```
kbarrette: Are these comments meant to be readable by the component implementor? I'm not sure if "EWB" is a recognizable term in that context, or if it's appropriate to discuss the graphql types

packages/spectral/src/serverTypes/index.ts
```diff
@@ -81,6 +82,9 @@ export interface Action {
   dynamicBranchInput?: string;
   perform: ActionPerformFunction;
   examplePayload?: unknown;
+  examplePerform?: ActionPerformFunction;
+  examplePerformSupport?: PerformSupport;
```
kbarrette: Are we OK with these two being the same type? They're the same now, but I semantically could be different.

kbarrette: Are we happy with these names? `performSupport = SAFE` means "it is safe to run perform", not totally sure "support" is the right word, but I also don't have a better suggestion

packages/spectral/src/serverTypes/convertComponent.test.ts
CaptainChemist: I was wondering this too. Claude said and it makes sense that for the expects in this file, we need to test the literal output that the backend expects so that if we decide to change around the string we can get a heads up that the backend will need the corresponding change too.

This is the same story over in the publishWire.spec.ts for all of the string expects. The case that's a bit fuzzier is in the `render({...})` blocks in that file where these values are used to control what the generator receives and so there are strings there but it could be the enums if we wanted but they were left as strings for consistency.

packages/spectral/src/generators/cniComponentManifest/index.ts
CaptainChemist: This is now called `PerformSupport` after I removed the experimental prefix. Is this okay or do we want `Inline` added? @kbarrette did you have plans one way or another on the backend?

packages/spectral/src/serverTypes/convertComponent.ts
CaptainChemist: Removed the experimental prefix after some discussion with the team.

packages/spectral/src/types/ActionDefinition.ts
CaptainChemist: That's a great point, I updated the comment to just call it the Prismatic UI and I removed the graphQL references.

packages/spectral/src/serverTypes/index.ts
CaptainChemist: I agree and as discussed made the following changes to `examplePerform` and `examplePerformSupport` to match the enum names:
 
Keep enums `SAFE`, `UNSAFE`, `NOT_ALLOWED` and change field names to safety related
Reads:
- `performSafety: SAFE | UNSAFE | NOT_ALLOWED` which means its safe, unsafe, or not allowed to invoke the perform
- `examplePerformSafety: SAFE | UNSAFE | NOT_ALLOWED` which means its safe, unsafe, or not allowed to invoke the example perform

packages/spectral/src/generators/cniComponentManifest/index.ts
CaptainChemist: This type has now been updated to `PerformSafety` to make it match better with the enums `SAFE`, `UNSAFE`, and `NOT_ALLOWED`.

kbarrette (approved):
Good to go to preview!

PR #498, opened 2026-07-13
Add support for inline action calling

CaptainChemist:
# What
This adds support on the spectral side for inline action calling which lets the integration builder invoke an action's perform (or mock examplePerform) from the Designer/EWB to populate live reference data.

# Changes
Author Facing type changes
- `PerformSafety` exposed along with a const companion which allows for authors to use either `PerformSafety.safe` or the string `"safe"` and get typechecking with both.
- Values are `"safe" | "notAllowed"` to match the backend graphQL enum type (`ActionPerformSafety`).

Publish-wire conversion
- `convertAction` wraps `examplePerform` via `createPerform` to add input cleaning and error handling and mirroring `perform`. Its destructured out of the `...action` spread so the raw function doesn't leak onto the wire when it isn't wrapped.
The published `Action` wire type declares `examplePerform` & both `examplePerformSafety` and `performSafety` fields.

Component manifest
- `ComponentManifestAction` adds the two `*Safety` scalars into the manifest so the frontend knows they are inline invokable.
- `examplePerform` is added to `ComponentManifestAction` type only. The generator doesn't emit a value or stub for it because its invoked by the published server action rather than the manifest.

CNI Component Manifest
- Threads the two `*Safety` flags through the graphQL fetch path so components fetched from the API surface them.

### DIFF
--- a/packages/spectral/src/generators/cniComponentManifest/types.ts
+++ b/packages/spectral/src/generators/cniComponentManifest/types.ts
@@ -1,6 +1,12 @@
-import type { Action, Component, DataSource, Trigger, TriggerPayload } from "../../serverTypes";
+import type {
+  Action,
+  Component,
+  DataSource,
+  PerformSafety,
+  Trigger,
+  TriggerPayload,
+} from "../../serverTypes";
 import type { ConfigVarResultCollection, Inputs } from "../../types";
-import type { PerformSafety } from "../../types/ActionDefinition";
 import type { TriggerResult } from "../../types/TriggerResult";
 
 export interface ComponentNode {
--- a/packages/spectral/src/generators/cniComponentManifest/types.ts
+++ b/packages/spectral/src/generators/cniComponentManifest/types.ts
@@ -1,5 +1,6 @@
 import type { Action, Component, DataSource, Trigger, TriggerPayload } from "../../serverTypes";
 import type { ConfigVarResultCollection, Inputs } from "../../types";
+import type { PerformSafety } from "../../types/ActionDefinition";
 import type { TriggerResult } from "../../types/TriggerResult";
 
 export interface ComponentNode {
@@ -30,6 +31,8 @@ export interface ActionNode {
     nodes: InputNode[];
   };
   examplePayload: string | null;
+  examplePerformSafety: PerformSafety | null;
+  performSafety: PerformSafety | null;
 }
 
 export interface ConnectionNode {
@@ -52,7 +55,10 @@ export interface InputNode {
   onPremiseControlled: boolean;
 }
 
-export type FormattedAction = Pick<Action, "key" | "display" | "inputs" | "examplePayload"> & {
+export type FormattedAction = Pick<
+  Action,
+  "key" | "display" | "inputs" | "examplePayload" | "examplePerformSafety" | "performSafety"
+> & {
   /** Emitted verbatim into the generated manifest; shape mirrors `examplePayload`. */
   outputSchema?: unknown;
 };
--- a/packages/spectral/src/generators/componentManifest/templates/actions/action.ts.ejs
+++ b/packages/spectral/src/generators/componentManifest/templates/actions/action.ts.ejs
@@ -28,4 +28,10 @@ export const <%= action.import %> = {
   <%_ if (action.outputSchema) { -%>
   outputSchema: <%- typeof action.outputSchema === 'string' ? action.outputSchema : JSON.stringify(action.outputSchema, null, 2) %>,
   <%_ } -%>
+  <%_ if (action.examplePerformSafety) { -%>
+  examplePerformSafety: "<%= action.examplePerformSafety %>",
+  <%_ } -%>
+  <%_ if (action.performSafety) { -%>
+  performSafety: "<%= action.performSafety %>",
+  <%_ } -%>
 } as const;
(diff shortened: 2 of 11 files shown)


packages/spectral/src/generators/cniComponentManifest/types.ts
```diff
@@ -52,7 +61,10 @@ export interface InputNode {
   onPremiseControlled: boolean;
 }
 
-export type FormattedAction = Pick<Action, "key" | "display" | "inputs" | "examplePayload"> & {
+export type FormattedAction = Pick<
+  Action,
+  "key" | "display" | "inputs" | "examplePayload" | "examplePerformSafety" | "performSafety"
```
robertmaloney, line 66: Should we just start omitting properties instead of expanding this everytime?

packages/spectral/src/generators/componentManifest/templates/actions/action.ts.ejs
```diff
@@ -28,4 +28,10 @@ export const <%= action.import %> = {
   <%_ if (action.outputSchema) { -%>
   outputSchema: <%- typeof action.outputSchema === 'string' ? action.outputSchema : JSON.stringify(action.outputSchema, null, 2) %>,
   <%_ } -%>
+  <%_ if (action.examplePerformSafety) { -%>
```
robertmaloney, line 31: Looks like perform is emitted so maybe examplePerform should (conditionally) be too?

packages/spectral/src/serverTypes/inlineActionCalling.publishWire.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+  });
+
+  it("C — real-perform SAFE flag carried; no separate example perform", () => {
+    const a = actionsByKey.runnablePerformAction;
+    expect(a.performSafety).toBe("SAFE");
+    expect("examplePerform" in a).toBe(false);
+    expect(a.examplePerformSafety).toBeUndefined();
+  });
+
+  it("D — real-perform notAllowed converts to NOT_ALLOWED; no example perform", () => {
```
robertmaloney: I think the mock action already uses `NOT_ALLOWED` (L71?) so this doesn't cover anything.

packages/spectral/src/types/ActionDefinition.ts
```diff
@@ -5,6 +5,17 @@ import type { ActionDisplayDefinition } from "./DisplayDefinition";
 import type { ConfigVarResultCollection, Inputs } from "./Inputs";
 import type { OutputSchema } from "./OutputSchema";
 
+/**
+ * Whether a perform is safe to invoke inline (to populate reference data in the Prismatic UI):
+ * `safe` to run as-is, `notAllowed` to opt out.
+ */
+export const PerformSafety = {
+  SAFE: "safe",
+  NOT_ALLOWED: "notAllowed",
```
robertmaloney, line 14: Is there a reason we want to allow friendly camelCase values? Seems we could simplify if this was just `SAFE: "SAFE"` etc

robertmaloney (approved):
Just nitpicks, overall lgtm!

packages/spectral/src/generators/cniComponentManifest/types.ts
CaptainChemist: Where Pick shines is that if new fields are added to `Action` they flow into `FormattedAction` by default silently. For example in this PR `Action` gained 3 fields but only 2 belong in the manifest. With `Omit` someone would have to remember to add it to the exclusion list.

To give a  related example of how this kind of issue has burned us, Lee has a prismatic commit 92e8a217bf for large data syncs. He had to add backend handling for `triggerResolverSupport` because in that case, the field was explicitly whitelisted on the spectral side but even then, the backend was serializing non-large data sync compatible triggers as `invalid` since that is the default. To address it, he had to add proper handling for that case on the backend. 

So that's all to say that even when you whitelist a new field in spectral, it can require backend changes to handle it properly, so I'm worried about whitelisting by default where fields the developer isn't necessarily thinking about gets green-lit automatically.

packages/spectral/src/types/ActionDefinition.ts
CaptainChemist: I did originally have this as `SAFE: "SAFE"` per our discussion of prioritizing developer UX in the form of consistency between the types here and the wire format you might see in Chrome dev tools. On this past Monday, Brian and Ryan pointed out that spectral has had a house convention of camelCase and that they wanted it here for consistency and so that's why I needed the casing translation here.

packages/spectral/src/generators/componentManifest/templates/actions/action.ts.ejs
CaptainChemist: My understanding of this is that we need `perform` here because its a dispatch stub that gets executed at runtime via `context.components`. `examplePerform` instead uses inline action calling via `fetchActionContent` so there isn't a consumer for `examplePerform` in the same way as `perform` in this context. Are there other uses for the `examplePerform` in this context that I'm missing?

packages/spectral/src/generators/cniComponentManifest/types.ts
robertmaloney: That's a fair call that this promotes an opt-in experience, sounds good!

packages/spectral/src/generators/componentManifest/templates/actions/action.ts.ejs
robertmaloney: Oh I misunderstood this to be a documentation thing, not a CNI thing. Carry on!

PR #502, opened 2026-07-15
Fix stable key and redirect urls leaking into component types

ryanwersal:
Adds type tests throughout including a new Exact helper which helps catch the `stableKey` class of error.

This also noticed that we re-exported the type utils so I've removed that as I couldn't find a downstream usage needing it.


bkegley (approved):
Looks good. As we've talked, encouragement to more explicitly distinguish component and integration authoring interfaces.

PR #513, opened 2026-08-10
Add userActivatedConnection for code-native integrations

jasoncomes:
Story: [SC-38384](https://app.shortcut.com/prismatic/story/38384/fe-declare-user-scoped-connections-from-code)

Lets a code-native author declare that a reusable connection is activated by **each person** who uses the integration, rather than once by the organization.

## `userActivatedConnection`

```ts
const userSlackConnection = userActivatedConnection({
  stableKey: "user-slack-connection",
});
```

It emits **the same definition as its siblings**. That is the design, not an oversight: scope is resolved server-side from the Scoped Config Variable that the stable key names, so the declaration carries no scope of its own and the wire shape needs no new field. The three helpers differ only in what they tell a reader - exactly as `customerActivatedConnection` and `organizationActivatedConnection` already do today.

Exported from both entry points (`integration.ts` uses an explicit named list, so code-native authors importing from `dist/integration` would not otherwise see it).

## Widened the scoped config var filters

Six sites used `OrganizationActivatedConnectionConfigVar` where they meant the whole `ScopedConfigVar` union. Worth knowing the history, because it wasn't a bad decision — it wasn't a decision at all:

- **2024-10-04** (`7451e1e`) - the union was born as `type ScopedConfigVar = OrganizationActivatedConnectionConfigVar`, a **one-member alias**. The two names were the same type, so writing either was identical.
- **2025-03-05** (`7cdf5ac`) - `customerActivatedConnection` landed and widened the union to two members. The filters weren't revisited. That same commit renamed the guard from `isOrganizationActivatedConnectionConfigVar` to `isConnectionScopedConfigVar`, acknowledging the new generality, but left its return type on the org member.

Nothing caught it because the members are structurally identical, so every filter kept working and `tsc` had nothing to complain about. **No behavior change here** - this just makes the filters say what they mean before the types ever diverge.

### DIFF
--- a/packages/spectral-test/src/ConfigVars.test-d.ts
+++ b/packages/spectral-test/src/ConfigVars.test-d.ts
@@ -18,6 +18,7 @@ expectAssignable<"A Connection" | "Ref Connection">(null as unknown as keyof Con
 expectAssignable<Connection>(null as unknown as ConfigVars["A Connection"]);
 expectAssignable<Connection>(null as unknown as ConfigVars["Ref Connection"]);
 expectAssignable<Connection>(null as unknown as ConfigVars["My Customer Connection"]);
+expectAssignable<Connection>(null as unknown as ConfigVars["My User Connection"]);
 
 expectAssignable<ConfigVar>({
   stableKey: "cni-oauth-connection",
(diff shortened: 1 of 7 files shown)


jasoncomes:
Closed by a branch rename, not intentionally. Superseded by #514 on `feature/user-scoped-connections` — same commit, same content.

PR #524, opened 2026-08-27
Fix structured object manifest input types

ryanwersal:
CNI manifests now rebuild structured and dynamic object inputs from the API’s flat field list, bringing them in line with the published manifests. The live smoke test also uncovered a bug in how choice models were queried and parsed, so those now follow the API’s JSON encoding. Snapshots and type checks cover the generated output and all supported input types.


bkegley (approved):
Thanks for adding the manifest type test. That alleviates my concerns around removing the explicit `undefined` properties from the manifest.

PR #525, opened 2026-08-27
Keep a user-activated connection on its user level page

jasoncomes:
## The problem

A code-native integration declaring a `userActivatedConnection` cannot be imported. The platform refuses it:

```
Error: definition: Required Config Var 'Your Dropbox Account' cannot use
user-level reusable connection 'dropbox-connection-baaaaaaaa'.
```

`convertConfigPages` removes scoped connections from a page's `elements`, and did so for user-activated connections too. That is right for a reusable connection – it is activated once, ahead of the wizard, and is not something a person fills in. It is wrong for this one.

A user-activated connection is activated by each person individually, and **the page it was declared on is the only record that it is user level**. The platform reads that from page membership (`build_ulc_config_var_mapping`) and has nothing else to go on. With the element removed, the connection was indistinguishable from ordinary instance configuration.

Nothing could supply the missing fact and there was no way to author around it, so this shape simply did not work from a code-native integration. Integrations built in the designer were unaffected, which is why it went unnoticed.

## The change

```diff
- .filter(([_key, value]) => !isConnectionScopedConfigVar(value))
+ .filter(([_key, value]) => !isConnectionReusable(value))
```

Whether a connection survives as an element comes down to one question: is it reusable, or does each person activate their own.

```ts
isConnectionReusable = (cv): cv is
  | CustomerActivatedConnectionConfigVar
  | OrganizationActivatedConnectionConfigVar =>
  isConnectionScopedConfigVar(cv) && !isUserScopedConnectionConfigVar(cv);
```

That is not a new idea in this codebase, which is the reason for the name. The manifest generator already emits a `ReusableConnection` returning `OrganizationActivatedConnectionConfigVar | CustomerActivatedConnectionConfigVar` – exactly the set this guard matches – so the guard narrows to that union rather than returning `boolean`, and the comments and the build error use the same word.

`isConnectionScopedConfigVar` deliberately still matches a user-activated connection, because it is one. `convertConfigVar` relies on that to emit the same `useScopedConfigVar` shape for every scoped kind. Narrowing it instead would push the special case somewhere worse.

The exception is only reachable on a user level page – the guard above already throws if a user-activated connection is declared anywhere else.

A kept element falls through to the existing default and emits `{ type: "configVar", value: key }`, which is the shape the platform reads. No new element type, no change to any other page.

## Emitted output

Before:

```yaml
configPages:
  - name: Your Dropbox
    userLevelConfigured: true
    elements: []
```

After:

```yaml
configPages:
  - name: Your Dropbox
    userLevelConfigured: true
    elements:
      - type: htmlElement
        value: Introduction
      - type: configVar
        value: Your Dropbox Account
```

## Verification

- 333 tests pass; `biome check` clean
- One test updated – it asserted the removal that caused this – and one added, covering that ordinary pages still drop reusable connections
- **Imported end to end against a live stack.** A code-native integration carrying a user-activated connection now imports, and the platform records it correctly:

```
estate-multi-simple  codeNative=True  integration.userLevelConfigured=True
   Your Dropbox Account | rcv.userLevelConfigured=True | orgOnly=False
                        | dropbox-connection-baaaaaaaa  USER
```

The same integration was refused before this change.

### DIFF
--- a/packages/spectral/src/types/ScopedConfigVars.ts
+++ b/packages/spectral/src/types/ScopedConfigVars.ts
@@ -19,9 +19,8 @@ export type OrganizationActivatedConnectionConfigVar = {
  * A connection each person activates for themselves.
  *
  * Carries its own `dataType` so it can be told apart from the organization- and
- * customer-activated kinds, which are otherwise structurally identical. See
- * `ConfigPageElement` for what that distinction buys. The convert layer rewrites it
- * to the shape the API expects, so this name never reaches the server.
+ * customer-activated kinds, which are otherwise identical. The convert layer rewrites
+ * it to the shape the API expects before it reaches the server.
  */
 export type UserActivatedConnectionConfigVar = {
   dataType: "userScopedConnection";
@@ -80,9 +79,8 @@ export const isConnectionScopedConfigVar = (cv: unknown): cv is ScopedConfigVar 
     return false;
   }
 
-  // Applied to both kinds: a declaration carrying a component reference or its own
-  // inputs is a connection definition rather than a reference to a Scoped Config
-  // Variable, whichever `dataType` it claims.
+  // A declaration carrying a component reference or its own inputs is a connection
+  // definition, not a reference to a Scoped Config Variable.
   return (
     !isConnectionDefinitionConfigVar(cv as ConfigVar) &&
     !isConnectionReferenceConfigVar(cv as ConfigVar)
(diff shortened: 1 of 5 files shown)


packages/spectral/src/types/ScopedConfigVars.ts
```diff
@@ -89,6 +89,20 @@ export const isConnectionScopedConfigVar = (cv: unknown): cv is ScopedConfigVar
...
+/**
+ * Whether this is a reusable connection: one the organization or the customer
+ * activates once, which every person then shares. True of every scoped kind
+ * except the user-activated one, the only connection an individual activates for
+ * themselves.
+ *
+ * This is the distinction that decides whether a connection survives as a config
+ * page element: a reusable one is stripped, because nobody fills it in.
+ */
+export const isConnectionReusable = (
```
robertmaloney: Technically a user scv is reusable too, so I think the name of this doesn't explain it's boundary accurately. Maybe just hoist the two checks to the filter you're using it in? Or maybe we can call this `isNonULCReusableConnection` or something.

robertmaloney (approved):

jpeters5392 (approved):

PR #526, opened 2026-08-31
Add @prismatic-io/eslint-plugin-spectral for section linting rules

pattra:
This PR adds a new package, `@prismatic-io/eslint-plugin-spectral`, which includes linting rules for using CNI sections properly. It also updates our provided eslint config to correctly point to it and references those rules. 

Note that these rules are already enforced at run-time, but linting provides an earlier guard to hopefully streamline the development cycle a bit. Rules, w/ more detail in `packages/eslint-plugin-spectral/README.md`:

| Rule | Catches |
| --- | --- |
| `@prismatic-io/spectral/no-nested-section` | A section opened while another is still open |
| `@prismatic-io/spectral/no-unclosed-section` | A section never closed, or a `sectionEnd` with nothing open |
| `@prismatic-io/spectral/section-label-match` | A `sectionEnd` label that does not match the open section |
| `@prismatic-io/spectral/no-section-outside-flow` | A section opened in a helper rather than in a flow handler |

These rules are all listed as errors, since they should ideally block further progression until addressed.

This PR also includes the release infra for the new plugin package. A new package was needed because new rules cannot be defined in config; we must define them elsewhere to fit with how eslintrc resolves them. 

I plan to release the v1 plugin manually first, then config. Once that is through, we should update prism to bump the config version so all newly created integrations include the new rules and check them at build time (we also may need to inform existing customers to do this bump as well).

### DIFF
--- a/packages/eslint-plugin-spectral/README.md
+++ b/packages/eslint-plugin-spectral/README.md
@@ -75,11 +75,14 @@ Inline callbacks within a handler are fine, since they are still visible in the 
 ```ts
 onExecution: async (context) => {
   for (const record of records) {
-    logger.section(record.id); // allowed
+    logger.section(record.id); // allowed by this rule
     logger.sectionEnd({ label: record.id });
   }
 },
 ```
+
+A section opened in a loop is still subject to the section limit, which `section-in-loop`
+reports separately.
 
 The handler names can be overridden:
 
@@ -91,6 +94,39 @@ Note that this rule only ever applies to code-native integrations. Connector `pe
 functions receive a plain `ActionLogger` with no `section` methods at all, so there is nothing
 for it to report there.
 
+### `section-in-loop`
+
+Advisory. A flow execution records at most 1000 sections. A `section()` opened in a loop creates
+one section per iteration, so a loop over a collection that grows with a customer's data can
+exhaust the limit. Past it the logs are still written, but are no longer grouped into sections.
+The runtime warns in the execution log when an execution reaches the limit; this rule reports the
+loop that gets you there before you deploy.
+
+```ts
+for (const record of records) {
+  logger.section(record.id); // reported
+  logger.sectionEnd({ label: record.id });
+}
+```
+
+Prefer one section per batch over one per record:
+
+```ts
+logger.section(`page ${page}`);
+for (const record of records) {
+  logger.info(record.id);
+}
+logger.sectionEnd({ label: `page ${page}` });
+```
+
+Array iteration methods count as loops too — `forEach`, `map`, `flatMap`, `filter`, `reduce`,
+`reduceRight`, `some`, and `every`.
+
+This is the one rule the recommended config sets to `warn` rather than `error`. It cannot know how
+many times a loop will run, so it reports every section opened in a loop, including loops that stay
+well under the limit. A loop over a handful of records is fine, and silencing the warning with
+`// eslint-disable-next-line` is a reasonable response.
+
 ## Limitations
 
 Each function body is analyzed on its own, in source order. That keeps the rules free of false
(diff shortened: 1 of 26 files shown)


packages/eslint-plugin-spectral/src/rules/noSectionOutsideFlow.ts
```diff
@@ -0,0 +1,78 @@
...
+      handlers.has(ancestor.key.value)
+    );
+  }
+
+  if (ancestor.type === "VariableDeclarator" && ancestor.id.type === "Identifier") {
+    return handlers.has(ancestor.id.name);
+  }
+
+  return ancestor.type === "FunctionDeclaration" && !!ancestor.id && handlers.has(ancestor.id.name);
+};
```
internetlee: I'm having a tough time following this (because I don't know the TSESTree structure, not because of your implementation). Is this also validating a case where `handlers=["onExecution", "fetchData"]` and `onExecution` calls `fetchData` within it?

I don't think we need to, but that seems to be the main way someone can get themselves a passing check, but be in an invalid runtime state. The hoops you have to do step through are a sufficient guard if it's not an easy thing to include or expensive for stepping through the tree for a deeply nested helper that gets stuffed into `handlers`. Likely best handled in docs on the site in examples/guidance for sections.

internetlee (approved):

packages/eslint-plugin-spectral/src/rules/noSectionOutsideFlow.ts
pattra: To answer the question -- no, it does not validate that case. This check is lexical, not call-based, and the handlers config is an intentional escape hatch, in case you have a real reason for putting logic in a function NOT named `onExecution`. 

It's true people can work around lint rules by modifying the code directly, and that's just outside of the scope of what we can do here. Our runtime checks and noisy logs are there for a reason and we will have appropriate documentation.

packages/eslint-plugin-spectral/src/rules/noSectionOutsideFlow.ts
```diff
@@ -0,0 +1,78 @@
+import type { TSESLint, TSESTree } from "@typescript-eslint/utils";
+import { createRule, withSectionCalls } from "../sections";
+
+type MessageIds = "outsideFlow";
+type Options = [{ handlers?: string[] }?];
+
+const DEFAULT_HANDLERS = [
+  "onExecution",
```
katrynmc: I thought we were only supporting onExecution to start? I think the remaining entries should be dropped. As written, the rule permits section calls in handlers where the logger types/runtime don’t support them.

packages/eslint-plugin-spectral/README.md
```diff
@@ -0,0 +1,140 @@
...
+section open.
+
+```ts
+logger.section("orders");
+logger.sectionEnd({ label: "order" }); // reported
+```
+
+### `no-section-outside-flow`
+
+Sections belong in a flow handler — `onExecution`, `onTrigger`, `onDeployTrigger`,
```
katrynmc: Only onExecution by my understanding

packages/spectral/src/testing.ts
katrynmc: Let's find a way to drop this any. 

Image

I think we can make this typesafe with the following:

```ts
type CodeNativeTestContext<TConfigVars extends ConfigVarResultCollection> = ActionContext<
  TConfigVars,
  Record<string, ComponentManifest["actions"]>,
  string[],
  CodeNativeActionLogger
>;

const createCodeNativeActionContext = <TConfigVars extends ConfigVarResultCollection>(
  context?: Partial<CodeNativeTestContext<TConfigVars>>,
): CodeNativeTestContext<TConfigVars> => {
  const actionContext = createActionContext<TConfigVars>(context);

  return {
    ...actionContext,
    logger: context?.logger ?? loggerMock(),
  };
};

```

Change context to:
```ts
context?: Partial<CodeNativeTestContext<TConfigVars>>;
```

And then realizedContext:
```ts
  const realizedContext = createCodeNativeActionContext({
    ...context,
    configVars: realizedConfigVars,
  });
```

and then you can drop the `any`. 

More broadly we should turn off `any` as separate work (not an ask, happy to do this myself).

PR #529, opened 2026-09-01
Restrict to org- and customer-mangaged SCVs in fetchConnectionStableKeys

kbarrette:


robertmaloney (approved):

packages/spectral/src/generators/cniComponentManifest/index.ts
```diff
@@ -260,7 +260,10 @@ export const fetchConnectionStableKeys = async ({
 
   const query = `
     query connectionStableKeysQuery($componentSelector: [ComponentSelector]!) {
-      scopedConfigVariables(connection_Component_In: $componentSelector) {
+      scopedConfigVariables(
+        connection_Component_In: $componentSelector
+        managedBy_In: ["org", "customer"]
```
jasoncomes, line 265: Lets also add `variableScope_In: ["org", "customer"]` to filter user connections so we only produce these records, `org-activated global, org-activated customer, customer-activated, and the build-only leaf`.

jasoncomes (approved):

PR #538, opened 2026-09-09
Add batchTrigger for authoring batched component triggers

internetlee:
Adds a component authoring parallel to the `batchFlowTrigger` for CNI for net new batching-only triggers. Consolidates the resolvers down into the perform functions by handling them at the convert layer. 

```
const syncOrders = batchTrigger({
  display: { label: "Sync Orders", description: "Fetches every order, one page at a time" },
  inputs: { connection: input({ label: "Connection", type: "connection", required: true }) },
  scheduleSupport: "required",
  batchConfig: { batchSize: 50, concurrentBatchLimit: 1 },
  perform: async (context, payload: TriggerPayload<{ cursor: string }>, params) => {
    const page = await fetchOrders(params.connection, payload.paginationState?.cursor);
    return {
      items: page.orders,
      paginationState: page.nextCursor ? { cursor: page.nextCursor } : null,
    };
  },
});
```


packages/spectral/src/types/TriggerPerformFunction.ts
```diff
@@ -10,7 +10,7 @@ export type TriggerPerformFunction<
   TConfigVars extends ConfigVarResultCollection,
   TAllowsBranching extends boolean | undefined,
   TResult extends TriggerResult<TAllowsBranching, TriggerPayload>,
-  TPaginationState extends Record<string, unknown> = Record<string, unknown>,
+  TPaginationState extends object = object,
```
pattra: This is in several places, but what do we gain by swapping `Record<string, unknown>` to `object`?

internetlee: @pattra 
To allow interfaces that don't require index signatures. 

```
interface OrderCursor {
  cursor: string;
}

batchTrigger<Order, OrderCursor>({ ... });
```

`OrderCursor is not assignable to Record<string, unknown>` here.

Mostly seemed a bit more discoverable to the component author. These are both fine though so maybe it's not actually much of a win.

`type OrderCursor = { order: string }`

```
interface OrderCursor {
  cursor: string;
  [key: string]: unknown;
}
```

Can be convinced it's not worth doing.

pattra: I guess I'm just scared of it breaking something we haven't anticipated down the line, but I don't personally have anything against it.

pattra (approved):

packages/spectral/src/types/BatchTriggerDefinition.ts
```diff
@@ -0,0 +1,175 @@
...
+import type { TriggerEventFunction } from "./TriggerEventFunction";
+import type { TriggerPayload } from "./TriggerPayload";
+import type { TriggerBaseResult } from "./TriggerResult";
+
+/**
+ * One page of records returned by a batched trigger's `perform`. Items are split into batches
+ * of the flow's batch size, one execution per batch. A non-null `paginationState` fetches the
+ * next page.
+ */
+export interface BatchTriggerBaseReturn<TItem, TPaginationState extends object = object> {
```
bkegley: I don't believe we need to be prescriptive on the shape of `TPaginationState` we could do `BatchTriggerBaseReturn<TItem, TPaginationState>` (defaulting to `unknown` which would then allow for a `string` representing the cursor if that's how someone wanted to paginate.

packages/spectral/src/types/TriggerPerformFunction.ts
bkegley: I [suggested](https://github.com/prismatic-io/spectral/pull/538#discussion_r4020075052) letting this generic fallback to `unknown` which we should support. That removes my uneasiness about typing something as `object` as that's pretty permissive and tends to throw a wrench into `infer`ing other types from it's value.

packages/spectral/src/types/BatchTriggerDefinition.ts
internetlee: I can get onboard with that. The platform currently expects a structured object (and guards on it, for better or worse). So I think here we should `object -> Record<string, unknown>` for now along with the thread above, and can relax that if we need.

PR #541, opened 2026-09-14
Add support for `IntegrationConfiguration` defining a configuration object instead of distinct config pages/variables

bkegley:
This allows defining CNI configuration as `IntegrationConfiguration` instead of distinct config pages and variables. This shift paves the way for authoring headless configuration experiences.

This changes the mental model of instance configuration from discrete config variables of different types assigned to pages in the config wizard to a "bag of JSON" with a schema to store and reference properties throughout an integration's triggers and flows.

There are a few high-level concepts to help realign the mental model:
1. Configuration is a modeled JSON store declared with either a zod or JSON schema.
2. Migrations are handled in a user-authored `init` function instead of relying on hidden, stateful transactions across integration versions indexed by stable key.
3. Third-party connections (both reusable connections and those specific to the published integration) are first-class primitives instead of special variables on a config page.


packages/spectral/src/serverTypes/convertIntegration.ts
```diff
@@ -1664,9 +1716,22 @@ const convertOnExecution =
...
     onExecution: ActionPerformFunction,
     componentRegistry: ComponentRegistry,
+    connectionNames?: ConnectionNameMap,
   ): ServerActionPerformFunction =>
   async (context, params) => {
-    const actionContext = createCNIContext(context, componentRegistry);
+    const baseContext = createCNIContext(context, componentRegistry);
+
+    // The runner supplies `configuration` already parsed. An unconfigured
+    // instance gets `{}` so a flow reading into it sees `undefined` for the
```
jpeters5392, line 1725: We should normalize this here and not rely on the runner behaving like this

jpeters5392 (approved):

packages/spectral/src/serverTypes/convertIntegration.ts
bkegley: This is already done at the db column level so we should be fine here.

PR #542, opened 2026-09-14
Add experimental flags spectral consumers can opt into

bkegley:
`spectral` consumers can turn a flag on by augmenting `Experimental` with the specified experimental flag. Internally, this allows `WithExperimentalFlag` to gate a type on it. Disabled resolves to `{}` by default, the identity for intersection, so a gated property is absent from a context rather than optional on it.

Nothing gates on a flag yet; the runtime is unaffected either way.

For consumer opt-in example see: https://github.com/prismatic-io/spectral/pull/542#discussion_r4008238421


packages/spectral-test/src/Experimental.test-d.ts
```diff
@@ -0,0 +1,32 @@
...
+  ? TFlags[TFlag] extends true
+    ? TEnabled
+    : TDisabled
+  : TDisabled;
+
+interface OptedIn {
+  someFlag: true;
+}
+
+type Context = { always: string } & WithExperimentalFlag<"someFlag", { gated: number }>;
```
jpeters5392, line 20: Is the idea here that we will export individual experimental flags that they can use?  Or will we have to tell them the flagName and flagType in order for them to use it?

bkegley: The `type Context ...` line is what internal usage looks like for flagging the type.

Consumer opt-in looks like this (which I'll update the description with):

```ts
declare module "@prismatic-io/spectral" {
  interface Experimental { someFlag: true }
}
```

which would then result in an additional required `gated: number` property on `Context`.

jpeters5392 (approved):

